PR #112, opened 2019-05-26
Retrieve code for AIC as figure 7

namn12:
need to implement previous code for aic. make axes start at 0.


namn12:
if we have a list of 40 cells, the cells that are root cells and at 30 or above (ie root cells at end list  - which live the longest of all root cells) have an EL that is near 0 and an NF that is near 0 and MSD that is at 1). This is for a 1 state fit on 2 state lineages.

namn12:
issue was that bern is instantiated as 1/numstates. in 1 state model this become 1. and for cells that die, their bern is 0, so the probability of being in this state of bern 1 is 0, which causes nans.

namn12:
i ran 100 of our 2 state lineages (same parameters as the paper) and the error bars are the stdev. @aarmey can you look at the formula we use calculate AIC in the `getAIC` function in `thmm.utils`? because our results here are not promising and wondering if it a misimplementation of the formula rather than an underlying model issue. @shak360 

image

namn12:
line 104, also copied below:

```
def getAIC(tHMMobj, LL):
    '''
    Gets the AIC values. Akaike Information Criterion, used for model selection and deals with the trade off
    between over-fitting and under-fitting.
    AIC = 2*k - 2 * log(LL) in which k is the number of free parameters and LL is the maximum of likelihood function.
    Minimum of AIC detremines the relatively better model.

    Args:
        ----------
        tHMMobj (obj): the tHMM class which has been built.
        LL (list): a list containing log-likelihood values of Normalizing Factors for each lineage.

    Returns:
        ----------
        AIC_value_holder_rel_0 (list): containing AIC values relative to 0 for each lineage.
        [numStates]*len(AIC_value_holder) (list): a list containing the # of states, repeated # of lineage times.
        AIC_degrees_of_freedom_holder (list): Contains all of the parameters for each of the lineages.


    Example Usage:

        from matplotlib.ticker import MaxNLocator

        x1val = []
        x2val = []
        yval = []
        for numState in range(3):
            tHMMobj = tHMM(X, numStates=numState, FOM='G') # build the tHMM class with X
            tHMMobj, NF, betas, gammas, LL = fit(tHMMobj, max_iter=100, verbose=False)
            AIC_value, numStates, deg = getAIC(tHMMobj, LL)
            x1val.append(numStates)
            x2val.append(deg)
            yval.append(AIC_value)

        fig = plt.figure(figsize=(10,10))
        ax1 = fig.add_subplot(111)
        ax1.scatter(xval, yval, marker='*', c='b', s=500, label='One state data/model')
        ax1.xaxis.set_major_locator(MaxNLocator(integer=True))
        ax1.grid(True, linestyle='--')
        ax1.set_xlabel('Number of States')
        ax1.set_ylabel('AIC Cost')
        title = ax1.set_title('Akaike Information Criterion')
        title.set_y(1.1)
        fig.subplots_adjust(top=1.3)

        ax2 = ax1.twiny()
        ax2.set_xticks([1]+ax1.get_xticks())
        ax2.set_xbound(ax1.get_xbound())
        ax2.set_xticklabels(x2val)
        ax2.set_xlabel('Number of parameters')

        ax1.legend()
        plt.rcParams.update({'font.size': 28})
        plt.show()
    '''
    numStates = tHMMobj.numStates
    AIC_value_holder = []
    AIC_degrees_of_freedom_holder = []
    for num in range(tHMMobj.numLineages):
        number_of_parameters = 0
        number_of_parameters += 1

        AIC_degrees_of_freedom = numStates**2 + numStates * number_of_parameters - 1
        AIC_degrees_of_freedom_holder.append(AIC_degrees_of_freedom)
        AIC_value = -2 * LL[num] + 2 * AIC_degrees_of_freedom
        AIC_value_holder.append(AIC_value)

    AIC_value_holder_rel_0 = AIC_value_holder - min(AIC_value_holder)  # this line is to make it so the minimum value is 0
    return(AIC_value_holder, [numStates] * len(AIC_value_holder), AIC_degrees_of_freedom_holder) # no longer returning relative to zero
```

aarmey:
All the files in `venv` need to be removed.

shak360:
Looks like my second axis for the number of parameters vs. the number of states isn't working either... Also that code doesn't make sense. Doesn't that for loop indicate that `number_of_parameters` is **always** 1?

```
 for num in range(tHMMobj.numLineages):
        number_of_parameters = 0
        number_of_parameters += 1
```

namn12:
> All the files in `venv` need to be removed.

done

lineage/figures/Fig_Gen.py
```diff
@@ -113,69 +109,92 @@ def Lineage_Length(T_MAS=500, T_2=100, reps=3, MASinitCells=[1], MASlocBern=[0.8
...
-            logging.info('accuracy: {}'.format(accuracy))
+        logging.info('X size: {}, masterLineage size: {}, subLineage2 size: {}'.format(len(X), len(masterLineage), len(subLineage2)))
+        
+        #Call function for AIC 
+        if AIC:
+            x1val = []
+            x2val = []
+            yval = []
+            for numState in range(numState_start, numState_end+1):
+                print(f'numState:{numState}')
```
aarmey: logging.info

lineage/blah.py
```diff
@@ -0,0 +1,39 @@
+EL 0.006591187292408536 0.006591187292408536 1.0
+EL 0.003826507378001198 0.003826507378001198 1.0
```
aarmey: What is this?

namn12: deleted. used it for debugging previously

namn12:
> Looks like my second axis for the number of parameters vs. the number of states isn't working either... Also that code doesn't make sense. Doesn't that for loop indicate that `number_of_parameters` is **always** 1?
> 
> ```
>  for num in range(tHMMobj.numLineages):
>         number_of_parameters = 0
>         number_of_parameters += 1
> ```

i fixed the thmm issue. took all but the second line out of the for loop

lineage/tHMM_utils.py
```diff
@@ -159,22 +159,17 @@ def getAIC(tHMMobj, LL):
...
-        elif tHMMobj.FOM == 'E':
-            number_of_parameters += 1
-
-        AIC_degrees_of_freedom = numStates**2 + numStates * number_of_parameters - 1
-        AIC_degrees_of_freedom_holder.append(AIC_degrees_of_freedom)
-        AIC_value = -2 * LL[num] + 2 * AIC_degrees_of_freedom
-        AIC_value_holder.append(AIC_value)
+        number_of_parameters += 1
+
+    AIC_degrees_of_freedom = numStates**2 + numStates * number_of_parameters - 1
```
shak360: Still doesn't make sense, sorry. Why is the number of parameters dependent on the number of lineages?

lineage/UpwardRecursion.py
```diff
@@ -2,6 +2,7 @@
 
 import numpy as np
 from .tHMM_utils import max_gen, get_gen, get_parents_for_level, get_daughters
+import pdb
```
aarmey: Don't need this, right?

lineage/figures/figure7.py
```diff
@@ -1,36 +1,42 @@
...
+    ax[0].set_ylabel('AIC Cost')
+    ax[0].set_title('Akaike Information Criterion')
+    #ax[0].set_ylim(1.1)
+    #ax[0].subplots_adjust(top=1.3)
+    ax1 = ax[0].twiny()
+    ax1.set_xticks(x1val)
+    ax1.set_xbound(ax1.get_xbound())
+    ax1.set_xticklabels(x2val)
+    ax1.set_xlabel('Number of parameters')
+    matplotlib.rcParams.update({'font.size': 28})
```
aarmey: Shouldn't set the font size for specific plots. This should be done globally.

lineage/tHMM_utils.py
shak360: There's a T matrix which is numStates squared.
There's pi which is just numStates.
Then for each state, we want to find a Bernoulli theta and exponential lambda (2), so numStates *2.

`AIC_degrees_of_freedom = numStates**2 + numStates * number_of_parameters` is fine, but `number_of_parameters` should be 2 **always**.

shak360: Actually, just add likelihood to the graph (another line plotting the likelihood values) unless it proves to be too difficult with the current way of how you store the data. 

More importantly the number of parameters thing has to be fixed; it should be 2, unless I'm misunderstanding something.

lineage/UpwardRecursion.py
namn12: done

lineage/tHMM_utils.py
namn12: > Still doesn't make sense, sorry. Why is the number of parameters dependent on the number of lineages?

i am not sure. originally, i just set the number of parameters equal to `numStates**2` but someone (i think yourself) added the extra `numStates * number_of_parameters` where `number_of_parameters` is equal to `numlineages` ....should we delete the latter i have never seen it in a literature but it is spefici to our application...

namn12: > There's a T matrix which is numStates squared.
> There's pi which is just numStates.
> Then for each state, we want to find a Bernoulli theta and exponential lambda (2), so numStates *2.
> 
> `AIC_degrees_of_freedom = numStates**2 + numStates * number_of_parameters` is fine, but `number_of_parameters` should be 2 **always**.

I agree lets just do AIC_degrees_of_freedom = numStates**2 + numStates * (E.shape[1]) - 1` although E is a list so i guess `len(E[0,:])` which is the number of variables  in the thmm

namn12:
i added `venv` to the gitignore so the folder wouldnt get pushed. can i resolve that conflict from the gitignore?

namn12:
image
final figure

Issue #117, opened 2019-06-10
Fill out author roles


aarmey:
The author contributions statement should list what each person did. Look [here](http://ijfs.ir/uploads/Pdfz/Author%20contributions2.pdf) for an example.

namn12:
@aarmey can you provide the names of the authors who performed the experiments at ohsu for cell phase data we will be using

aarmey:
Don't worry about adding them for the moment. I'll have to check with them.

PR #126, opened 2019-06-22
Author contributions

namn12:


namn12:
@aarmey the authorline is based on what is currently in the manuscript as it is. after evaluating the manuscript in current form, nearly everything presented (ie all the data, model, and writing) was performed in substantial form by SV, NKN, and AW. however, AF and AL have done a lot of work for the project that did not see its way to the manuscript in its current form. likewise, most of the work performed by FM on thmm is not directly presented in the manuscript in its current form. though i know FM has been working substantially on the project and she has results that will soon be a part of the paper (ie cell cycle phase, etc). there is a lot at play and i dont think this is a finalized authorline (though the contribution descriptions should be relatively robust in the current form) and i decided to keep everyone on for the time being with understanding that things can possibly change. i should also note that i consulted the first author @shak360 before i made this decision. we acknowledge that you have the final say in the authorline. let us know what you think and i am happy to talk more about this.

shak360:
#117

aarmey:
Also you should use some consistent rule for initials ordering (just so it's not random). Say alphabetical?

namn12:
> Also you should use some consistent rule for initials ordering (just so it's not random). Say alphabetical?

done

Issue #118, opened 2019-06-10
Reorganize methods


aarmey:
I think the methods should start with the core framework of the tHMM, then the emissions/observations definition should be a smaller subsection afterwards. While we spent some time implementing the one state model, it's a trivial part of the larger tHMM and so I don't think it needs to be separately described in the methods.

namn12:
PR 129

Issue #116, opened 2019-06-08
Introduction paragraph on prior work


aarmey:
We should have a paragraph on prior applications of tHMMs since there are one or two. I would also mention linkage analysis here, and analysis of incomplete penetrance in family trees.

shak360:
**Are linkage studies boring?**
https://www.nature.com/articles/ng0793-213
_"Most important, using linkage as a foundation, the search for actual genes by positional cloning will accelerate and become more accurate. The eventual discovery of the ataxia genes will lead to better speculations about pathogenesis and ultimately to much needed treatment and prevention. Will these genes show deletions, duplications, point mutations, triplet repeat expansions, imprinting and anticipation?"_

**Genetics of Type 1A Diabetes**
https://www.nejm.org/doi/full/10.1056/NEJMra0808284?url_ver=Z39.88-2003&rfr_id=ori:rid:crossref.org&rfr_dat=cr_pub%3dpubmed
_ A disease allele that occurs frequently in the population and that has a large effect on disease risk is unlikely to exist. At the opposite end of the spectrum, a disease allele that is rare and has a small effect size is likely to exist but is unlikely to be found — and such alleles would be of limited public health interest. In general, linkage studies are most effective in disorders in which disease
alleles are anticipated to have a large effect size but occur infrequently. Association studies are most effective for the detection of alleles that occur frequently but have a small effect size. These are general trends, and there are no specific boundaries in efficacy between the two approaches._

shak360:
**Deciphering Latent Growth-States from Cellular Lineage Trees**
https://www.biorxiv.org/content/10.1101/488981v1
_"Then, we propose an estimation algorithm of the latent states from lineage trees based on an estimation-maximization (EM) algorithm, which we call Lineage EM algorithm (LEM). We verify the effectiveness of LEM by using synthetic data. Finally, we apply LEM to a lineage tree of E. coli, and identify a latent three-dimensional continuous state, one component of which encodes the information on the inheritance of the states and the replicative capabilities over generations. The inferred dynamics suggests that the homeostasis and heterogeneity of the division times are controlled with multiple time scales."_

**Modelling Stem Cells Lineages with Markov Trees**
https://link.springer.com/chapter/10.1007/978-3-642-04031-3_21
[...]

PR #129, opened 2019-07-06
Intro and Method update (#118 and #116)

namn12:
@aarmey Can you see if this addresses the points on issues #118 and #116 . Shakthi laid foundation for the intro and I just reorganized and edited where I saw fit. Method I didn't edit much because the thmm description was already present (I just made the methods begin with it) and I edited the last section of the methods to describe the emissions (rather than 1 state model). Let us know what you think and how we should move forward.


namn12:
great thanks

PR #130, opened 2019-07-08
More intro

shak360:
Went through and did a thorough sweep of the grammar checklist: clarity, punctuation, syntax, usage, etc. Tried to make some of the points clearer. Nothing was moved but several words/sentences are changed. Fit for the New Yorker, but will settle for Nature. Also all citations are moved to be before the periods, commas, and semicolons in the introduction.

For the methods, I was hoping we could move all equation/mathematics to the supplement and keep just one equation in the actual methods section and immediately move to the results. I want to do this because understanding the results requires little use of the mathematical methods but this is just a suggestion. I've seen several Cell/Nature papers do this so I thought it'd be cool and increase the readability, but its just a thought.


manuscript/01_introduction.md
```diff
@@ -34,48 +34,45 @@ csl: ./common/templates/nature.csl
...
... diagnosed with over 609,000 of those cases projected to be fatal [@Smith]. One of the primary treatments of cancer consists of chemotherapy, mainly targeted therapies, whereby patients are administered drugs that destroy highly-prolific cells to stall cancer growth or eliminate the tumor. Long-term therapeutic efficacy, however, varies significantly due to the vast heterogeneity in intratumor response to therapy [@DiMaio; @DeRoock]. Cell variability in drug response can originate from cell-intrinsic factors, such as genomic alterations (i.e., altered nucleotide excision repair, telomere maintenance, and copy-number variation), and cell-extrinsic factors such as spatial variability in the surrounding vasculature and environmental stressors causing DNA promoter mutations [@Feinberg; @Falkenberg; @Inde]. Current drug-screening protocols involve subjecting studied cancer cell lines to variable drug doses and evaluating therapy performance by examining (1) the percent of cells eliminated and (2) the dose required for half-maximal effect [@Chauvin; @Zhang_L]. These metrics, however, provide population averaged tumor responses to therapy that fail to consider subpopulation heterogeneity.
```
aarmey: 1.7 million cases is just for the US, right?

shak360: Yes, just the US; will change. What do you think about the suggestion for the methods Dr. Meyer  @aarmey? Can we make the methods not as intense (mathematically) and leave a lot of it in the supplements/appendix?

aarmey: That sounds reasonable.

PR #133, opened 2019-07-09
broken notebook for some reason

shak360:
I was fixing a weird bug where the notebook was saved as an HTML file for some reason. Should be fixed now; this is a trivial PR.


shak360:
Farnaz is working on fixing the failing tests in a different branch.

PR #149, opened 2019-07-25
refactoring

shak360:
Synthetic lineage generation is completely changed. Emissions handling is completely changed. tHMM is changed to accommodate these changes.
Main differences:
- cells are just objects that hold observations defined by a user 
- user defined emissions in StateDistribution file
- the full tree prior to pruning is now accessible if needed
- changing emissions only requires changing the StateDistribution file
- tree is now inherently heterogeneous (tree is first defined through its states via the pi and transition matrix probabilities)
- no such thing as a tHMM paramlist or a Force of Mortality; these are now both encoded in the state distribution objects

Things to do still:
- clean up guide to code notebook (Farnaz and I can do this)
- clean up tHMM.py (the E[state] initialization should come from StateDistribution.py (I'll change this)
- complete unit tests for (I can do this)
- - LineageTree
- - tHMM
- change figure code, take the time now to use individual plotting configurations for each figure (Nikan and I can do this, figure code should be a lot smoother to write now.)
- linting

For a review, we suggest one looks at CellVar.py, LineageTree.py, StateDistribution.py, and then Guide_to_tHMM.ipynb in that order. Note that some functions are still hacked together and aren't in the correct style of variable naming I'd like them to be. One aspect of tHMM is failing (the sum of the gamma probabilities isn't always 1 for all the cells, sometimes 0) but otherwise it works and reaches above 85% accuracy on full lineages. It does just as well on pruned lineages, so long as the transition matrix is appropriately designed (the resistant to susceptible transition is low). @Farnazmdi add any comments you think I missed out on.

Tests should be passing. Building is failing due to the figures still having old code. Will change.


lineage/LineageTree.py
```diff
@@ -0,0 +1,265 @@
...
+            seperator = ', '
+            s2 = seperator.join(s_list)
+            s3 = ".\n This UNpruned tree has {} cells in total".format(len(self.full_lin_list))
+            return s1 + s2 + s3
+
+# tools for traversing trees
+
+
+def tree_recursion(cell, subtree):
+    """ a recurssive function that traverses upwards from the leaf to the root. """
```
aarmey: Spell check.

lineage/LineageTree.py
```diff
@@ -0,0 +1,265 @@
+""" This file contains the LineageTree class. """
+
+from .CellVar import CellVar
+from .StateDistribution import prune_rule
+
+import scipy.stats as sp
+import numpy as np
```
aarmey: External imports go before internal ones.

lineage/StateDistribution.py
```diff
@@ -0,0 +1,184 @@
...
+
+# Because parameter estimation requires that estimators be written or imported, the user should be able to provide
+# estimators that can solve for the parameters that describe the distributions. We provide some estimators below as an example.
+# Their use in the ObservationEmission class is shown in the estimator class method. User must take care to define estimators that
+# can handle the case where the list of observations is empty.
+
+
+def report_time(cell):
+    """ Given any cell in the lineage, this function walks through the cell's ancestors and return how long it has taken so far. """
+    list_parents = [cell]
```
namn12: bookmark

lineage/StateDistribution.py
```diff
@@ -0,0 +1,184 @@
...
+        if cells._isRootParent():
+            taus.append(cells.obs[1])
+            break
+        elif cells.parent not in list_parents:
+            list_parents.append(cells.parent)
+            taus.append(cells.parent.obs[1])
+        summation = sum(taus)
+        return summation
+
+<<<<<<< HEAD
```
namn12: delete ?

Farnazmdi: I pushed my commits the other day! Seems like it hasn't been updated!!!

shak360:
Yeah IDK why there are so many mmerge issues. I fixed a lot last time and made some changes to the code. For example, I commited changes in:

https://github.com/meyer-lab/lineage-growth/pull/149/commits/72b833f6d1d9b4b27852420db2b2cebd602a2cbb

but for some reason, they're being reapplied in :

https://github.com/meyer-lab/lineage-growth/pull/149/commits/3ba1b2c02e4030973de26cd0748c0ead42233ec2

lineage/CellVar.py
```diff
@@ -0,0 +1,106 @@
...
+        temp = []
+        if self.left is not None:
+            temp.append(self.left)
+        if self.right is not None:
+            temp.append(self.right)
+        return temp
+
+    def __repr__(self):
+        if hasattr(self, 'obs'):
+            return "Generation: {}, State: {}, Observation: {}".format(self.gen, self.state, self.obs)
```
namn12: what is self.obs ? where is it defined ?

shak360: it doesn't need to be defined upon instantiation. All cells are empty (without observations) in the beginning. We then fill their observations in when the state distributions are known. This is done in LineageTree.py.

lineage/CellVar.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def __str__(self):
+        if hasattr(self, 'obs'):
+            return "Generation: {}, State: {}, Observation: {}".format(self.gen, self.state, self.obs)
+        else:
+            return "Generation: {}, State: {}, Observation: {}".format(self.gen, self.state, "This cell has no observations to report.")
+
+
+def _double(parent_state, T):
+    """ Function that essentially rolls two of the same loaded dice given a state that determines the row of the transition matrix. The results of the roll of the loaded dice are two new states that are returned. """
```
namn12: what are the two same loaded die you are referring to and what are the _two new states that are returned_ ?

namn12: i see youre initializing the T for this cell.

namn12: actually, can u explain this for me

shak360: Drawing samples from a multinomial distribution can be thought of as rolling a loaded die. It's the same way as drawing samples from a bernoulli distribution can be thought of as flipping a loaded coin. Each row of the T matrix represents a loaded die where the "loading" is the probability that the "die" lands on that state. We roll this loaded die (sample the multinomial distribution twice) when considering what the states are going to be for the two daughter cells from a parent cell of state_j. We go to the j_th row, use that row to load the die (to tell the multinomial distribution the probability of each state), roll the die (sample the distribution) and the two results we get are the states of the two daughters.

shak360:
@Farnazmdi the merge errors aren't being fixed correctly. You need to choose which commit you want to keep (either HEAD or MASTER). You can't delete the headings simply. You have to decide which block of code you want to keep.

Farnazmdi:
Oh, I see...
The guide was broke and even copying to a new script was not working,
I guess I made it worse.. do you have a correct version in you local? @shak360

shak360:
@Farnazmdi 
image
So in this example, say you wanna keep everything that's coming from HEAD. So what you do is delete lines 311-313, 433-435. Then you delete the lines 309 and 431. This process let's you pick what blocks of code you want to keep. 

If you don't want things from HEAD, then you delete lines 309-311 and 431-433. You then delete the lines 313, and 435. 

Note in both cases the headings are deleted at the end, as is the line with the equal signs. But you need to ensure only one block of code remains in these merge conflicts.

lineage/LineageTree.py
```diff
@@ -0,0 +1,285 @@
...
+                self.full_lin_list.append(left_cell)  # add daughters to the list of cells
+                self.full_lin_list.append(right_cell)
+
+            if len(self.full_lin_list) >= self.desired_num_cells:
+                break
+
+        return self.full_lin_list
+
+    def _prune_lineage(self):
+        """ This function removes those cells that are intended to be remove from the full binary tree based on emissions.
```
namn12: 'removed' spell typo. this all looks amazing

lineage/CellVar.py
shak360: lmk if this makes sense or not, or should we add more documentation

shak360:
> Oh, I see...
> The guide was broke and even copying to a new script was not working,
> I guess I made it worse.. do you have a correct version in you local? @shak360

I'm just going to revert back a commit, and we can fix this later. Use `git revert HEAD` to go back one commit, if the commit is already pushed.

lineage/LineageTree.py
```diff
@@ -0,0 +1,285 @@
...
+
+def tree_recursion(cell, subtree):
+    """ a recursive function that traverses upwards from the leaf to the root. """
+    if cell._isLeaf():
+        return
+    subtree.append(cell.left)
+    subtree.append(cell.right)
+    tree_recursion(cell.left, subtree)
+    tree_recursion(cell.right, subtree)
+    return
```
namn12: shouldnt this return subtree?

shak360: nah

Farnazmdi:
Still notebook doesn't load for me in the Github

lineage/LineageTree.py
shak360: it can, but since the subtree is already an input argument, it will be a mutable variable even after the function finished

shak360:
Yea it had the merge conflicts because I reverted them. I fixed the merge conflicts I think. Should be working now.

lineage/DownwardRecursion.py
```diff
@@ -1,36 +1,52 @@
...
+        lineage = lineageObj.output_lineage
+        gamma_array = np.zeros((len(lineage), numStates))
+
         gamma_array[0, :] = betas[num][0, :]
         assert np.isclose(np.sum(gamma_array[0]), 1.)
         gammas.append(gamma_array)
 
+        for num, lineageObj in enumerate(tHMMobj.X):  # for each lineage in our Population
+            gammas_0_row_sum = np.sum(gammas[num][0])
+            assert np.isclose(gammas_0_row_sum, 1.)
```
namn12: so this is failing right, can you make an issue, and we can address it after the merge if it still lingers. might be an initialization thing where depending on the initial thmmm params it can not converge and fail

shak360: Yea basically for the most part it works. It's just some are summing to 0, most of them are summing to 1. I feel like it might be the leaves? but I'm not sure.

lineage/tHMM.py
```diff
@@ -1,15 +1,22 @@
...
 import numpy as np
-import scipy.stats as sp
-from .Lineage_utils import get_numLineages, init_Population
+from .StateDistribution import StateDistribution, tHMM_E_init
 from .tHMM_utils import max_gen, get_gen
 
+class estimate:
+    def __init__(self, numStates):
+        self.numStates = numStates
+        self.pi = np.ones((numStates)) / numStates
```
namn12: what is the relationsihp between this estimate class and the state_distribution.py parameters

namn12:
> Still notebook doesn't load for me in the Github

i cant view it either

lineage/tHMM.py
```diff
@@ -1,15 +1,22 @@
 """ This file holds the parameters of our tHMM in the tHMM class. """
 
 import numpy as np
-import scipy.stats as sp
-from .Lineage_utils import get_numLineages, init_Population
+from .StateDistribution import StateDistribution, tHMM_E_init
 from .tHMM_utils import max_gen, get_gen
 
+class estimate:
```
namn12: so u have to call the estimate class (which holds the thmm class) for every cell in the lineage right?

lineage/figures/Fig_Gen.py
```diff
@@ -183,7 +185,12 @@ def Lineage_Length(T_MAS=500, T_2=100, reps=50, MASinitCells=[1], MASlocBern=[0.
     return data
 
 
+<<<<<<< HEAD
+def Lineages_per_Population_Figure(lineage_start=1, lineage_end=2, numStates=2, T_MAS=500, T_2=100, reps=1, MASinitCells=[1], MASlocBern=[0.8], MASbeta=[
+                                   80], initCells2=[1], locBern2=[0.999], beta2=[20], max_lin_length=300, min_lin_length=5, FOM='E', verbose=True, switchT=True):
+=======
 def Lineages_per_Population_Figure(lineage_start=1, lineage_end=2, numStates=2, T_MAS=500, T_2=100, reps=1, MASinitCells=[1], MASlocBern=[0.8], MASbeta=[80], initCells2=[1], locBern2=[0.999], beta2=[20], max_lin_length=300, min_lin_length=5, FOM='E', switchT=True):
+>>>>>>> master
```
namn12: im sure u know , merge conflict

lineage/tHMM.py
shak360: estimate class is restricted to tracking the pi, T, and E _estimates_. the StateDistribution object  as well as the LineageTree object holds the _true_ values.

lineage/tHMM.py
shak360: No we instantiate the class once when the tHMM instantiation is called.

shak360: the tHMM class holds one estimate class. The estimate class gets changed as thee various tHMM related functions are changed. It's similar to `paramlist` from before.

lineage/tests/test_StateDistribution.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def test_bernoulli_estimator(self):
+        """ blah """
+        bern_obs = sp.bernoulli.rvs(p=0.9, size=1000)  # bernoulli observations
+        self.assertTrue(0.88 <= bernoulli_estimator(bern_obs) <= 0.92)
+
+    def test_exponential_estimator(self):
+        """ blah """
+        exp_obs = sp.expon.rvs(scale=50, size=1000)  # exponential observations
+        self.assertTrue(45 <= exponential_estimator(exp_obs) <= 55)  # +/- 5 of beta
```
namn12: how did you decided upon 5 and .02 for the bern as the up/down boundaries

shak360:
https://github.com/meyer-lab/lineage-growth/blob/new_approach/Guide_to_tHMM.ipynb the notebook should be running. I fixed all the merge conflicts. Try again after pulling maybe idk

lineage/tests/test_StateDistribution.py
shak360: These were just the ones that caused the tests to fail the least.

lineage/figures/Fig_Gen.py
shak360: Yea we still have to rewrite the figure stuff, which should be cleaner to do now.

namn12:
a cool experiment would be to compare the gamma estimates with the same lineage tree pruned and unpruned to measure the survivorship bias in parameter estimation.

shak360:
> a cool experiment would be to compare the gamma estimates with the same lineage tree pruned and unpruned to measure the survivorship bias in parameter estimation.

exactly what i was thinking :))))

namn12:
> > a cool experiment would be to compare the gamma estimates with the same lineage tree pruned and unpruned to measure the survivorship bias in parameter estimation.

and exactly what you did,....had to scroll down a more in the notebook

namn12:
at the very last cell of your notebook the paragraph states: `The point to be considered here is that transition from 1 to 3 or otherwise is not possible so the probability of these transitions are zero, and most likely the initial cells are in susciptible state.\ $\textbf{State 0}$: susciptible\ $\textbf{State 1}$: middle state\ $\textbf{State 2}$: resistant`

lineage/CellVar.py
namn12: ah, use that row to roll the die is based on what state you are in determines the probabilities of transitioning ie of the die showing a particular face

namn12: `are the two new states` **of the daughter cells** that are returned is what you mean right.

lineage/LineageTree.py
namn12: so its global can u point me to how this is so

shak360:
> at the very last cell of your notebook the paragraph states: `The point to be considered here is that transition from 1 to 3 or otherwise is not possible so the probability of these transitions are zero, and most likely the initial cells are in susciptible state.\ $\textbf{State 0}$: susciptible\ $\textbf{State 1}$: middle state\ $\textbf{State 2}$: resistant`
> can you print out the average transition matrices for cells in states 1,2,3 ? how do you want to show that transitions from the 1 and 3 states are near 0? and can you print out the average p value for all cells? i assume we have to print out average values to get the gauge of the population thmm characterisitics.

There is only one pi, T, and E per tHMM not per cell. We estimate this pi T and E through the same way as before.

lineage/DownwardRecursion.py
namn12: it was the leaves last time this happend. and it was cuz their initialization params were too extreme

namn12:
> > at the very last cell of your notebook the paragraph states: `The point to be considered here is that transition from 1 to 3 or otherwise is not possible so the probability of these transitions are zero, and most likely the initial cells are in susciptible state.\ $\textbf{State 0}$: susciptible\ $\textbf{State 1}$: middle state\ $\textbf{State 2}$: resistant`
> > can you print out the average transition matrices for cells in states 1,2,3 ? how do you want to show that transitions from the 1 and 3 states are near 0? and can you print out the average p value for all cells? i assume we have to print out average values to get the gauge of the population thmm characterisitics.
> 
> There is only one pi, T, and E per tHMM not per cell. We estimate this pi T and E through the same way as before. The estimated pi T and E are stoerd in the thMM.estimate object.

i see now. the roll the die T in cellvar was to determined the state of the daughters. nice. t e and pi are stored the same from what i understand

namn12:
i approved , but why is it breaking ?

shak360:
> i approved , but why is it breaking ?

None of the figure code is changed. Also I haven't written tests for tHMM. Figured we could take the chance now to write tests the same way HMMlearn did.

shak360:
We could definitely advance without writing tests for tHMM, but I the figure code be changed I think, before we move on. We can take the time to split it up better (have each figure's code on separate files).

aarmey:
If you prefer we could comment out each figure's code so this builds, review what's here, merge, and then rebuild the tests/figures on separate branches. That might break all this work up a bit.

namn12:
> If you prefer we could comment out each figure's code so this builds, review what's here, merge, and then rebuild the tests/figures on separate branches. That might break all this work up a bit.

i agree

shak360:
Yea I think that's a good idea. Commenting out figure code and working on the tests/figures later would be easier.

shak360:
OK I think all the major issues are fixed now. The figure code is removed and the test files are there, just empty. This pull request is ready for a review Dr. Meyer @aarmey . The only issue with it is linting I think, according to Jenkins. I think it's maybe due to our gratuitous use of the `_` when naming things in classes, which is my fault but I wanted to use it so the user knows what functions to touch and what not to. The guide to code is also fully running now; the only thing that breaks in the guide to the code is that the gamma estimator fails sometimes. I find that simply running that cell block again and again fixes it. Again it might help to do CellVar -> LineageTree -> StateDistribution -> Guide_to_tHMM in that order. I also took the chance in the merge errors to make some optimizations related to storing generations of the tree, which you might see in LineageTree.

lineage/LineageTree.py
shak360: the thing that we append to (`subtree`) is also an argument of the function

see https://runestone.academy/runestone/books/published/thinkcspy/Lists/UsingListsasParameters.html

lineage/StateDistribution.py
shak360: it adds the times of the cells in a branch in a tree

given any cell, it just walks up the tree until it hits the root, adding the lifetimes along the way

lineage/tests/test_StateDistribution.py
```diff
@@ -0,0 +1,29 @@
...
+from ..StateDistribution import StateDistribution, bernoulli_estimator, exponential_estimator, gamma_estimator, report_time
+from ..LineageTree import LineageTree
+from ..CellVar import CellVar as c
+
+
+class TestModel(unittest.TestCase):
+    """Here are the unit tests."""
+
+    def test_bernoulli_estimator(self):
+        """ blah """
```
aarmey: Fill in real docstrings.

lineage/tests/test_CellVar.py
```diff
@@ -0,0 +1,127 @@
...
+        right_state = 1
+
+        cell_left = c(state=left_state, left=None, right=None, parent=None, gen=1)
+        cell_right = c(state=right_state, left=None, right=None, parent=None, gen=1)
+
+        self.assertTrue(cell_left.state == 0)
+        self.assertTrue(cell_right.state == 1)
+
+    def test_cell_divide(self):
+        """ blah """
```
aarmey: Real docstrings.

lineage/StateDistribution.py
```diff
@@ -0,0 +1,183 @@
...
+        self.bern_p = bern_p
+        self.expon_scale_beta = expon_scale_beta
+        self.gamma_a = gamma_a
+        self.gamma_scale = gamma_scale
+
+    def rvs(self, size):  # user has to identify what the multivariate (or univariate if he or she so chooses) random variable looks like
+        """ User-defined way of calculating a random variable given the parameters of the state stored in that observation's object. """
+        # {
+        bern_obs = sp.bernoulli.rvs(p=self.bern_p, size=size)  # bernoulli observations
+        exp_obs = sp.expon.rvs(scale=self.expon_scale_beta, size=size)  # exponential observations
```
aarmey: Remind me why there's both exponential and gamma here? Shouldn't the lifetimes be only one or the other?

lineage/LineageTree.py
aarmey: So this is filling in the `subtree` variable in place?

lineage/DownwardRecursion.py
```diff
@@ -50,4 +57,5 @@ def get_nonroot_gammas(tHMMobj, gammas, betas):
                         assert np.all(gammas[num][0, :] == betas[num][0, :])
 
     for _, gg in enumerate(gammas):
-        assert np.allclose(np.sum(gg, axis=1), 1.)
+        #assert np.allclose(np.sum(gg, axis=1), 1.)
```
aarmey: What's going on here?

lineage/StateDistribution.py
shak360: It was just a proof of concept sort of thing. I wanted to show that the number of observation distributions (emission) is arbitrary. I wanted to show that it can be any number of observed things (if that makes sense...)

lineage/DownwardRecursion.py
shak360: It fails in some runs and I'm not sure why, so I've just commented it out for now. Will ping as a new issue and work in a later branch.

lineage/LineageTree.py
shak360: Yep. Most of tHMM works by modifying these large lists of `Nxk` arrays in place (like the `betas`, `gammas`, `NF`, etc). Those are just lists, where each list item is an `Nxk` (`N` is the number of cells in that lineage, `k` is the desired amount of states). These arrays get modified heavily over the analysis.

 This is just a smaller example of that where `subtree` is a growing list of cells that are the subtree of a cell.

PR #156, opened 2019-08-09
Testing

Farnazmdi:
Added some tests for `LineageTree` and `StateDistribution`, and more docstring to the files. 
I am not sure about the warnings though.


Farnazmdi:
Should I just "enter" when a line gets too long?!

shak360:
no we can use autopep8 to fix those linting errors @Farnazmdi 

`autopep8 -i -a -r --max-line-length 200 lineage`

input the above into the terminal while in the lineage repository and enter

shak360:
@aarmey Linting is done and the tests are running successfully. IDK why the reviews become stale after changes. I don't recall it doing that earlier.

aarmey:
I'll change that setting. It's not ideal either way, but not annoying is better.

PR #161, opened 2019-08-20
Accuracy asm

aarmey:
I don't think the problem is (only) in BW. Look at comments below.


lineage/UpwardRecursion.py
```diff
@@ -44,17 +42,14 @@ def get_leaf_Normalizing_Factors(tHMMobj):
...
-                temp_sum_holder.append(joint_prob)  # append the joint probability to be summed
 
-            marg_prob = sum(temp_sum_holder)  # law of total probability
+            # P(x_n = x , z_n = k) = P(x_n = x | z_n = k) * P(z_n = k)
+            # this product is the joint probability
             # P(x_n = x) = sum_k ( P(x_n = x , z_n = k) )
             # the sum of the joint probabilities is the marginal probability
-            NF_array[leaf_cell_idx] = marg_prob  # each leaf is now intialized
+            NF_array[leaf_cell_idx] = np.sum(MSD_array[leaf_cell_idx, :] * EL_array[leaf_cell_idx, :])  # def of conditional prob
+            assert NF_array[leaf_cell_idx] > 0.0
```
aarmey: I'm pretty confident this should be true, since the likelihood of the tree comes from the sum of the log of these values.

lineage/tests/test_BaumWelch.py
```diff
@@ -0,0 +1,57 @@
...
+        E = [state_obj0, state_obj1]
+        
+        num = 10000
+
+        # Using an unpruned lineage to avoid unforseen issues
+        X = LineageTree(pi, T, E, num, prune_boolean=False)
+
+        tHMMobj = tHMM([X], numStates=2)  # build the tHMM class with X
+
+        LLbefore = calculate_log_likelihood(tHMMobj, get_leaf_Normalizing_Factors(tHMMobj))
```
aarmey: Both LL that I calculate are coming out to be -inf, even before running fitting.

lineage/UpwardRecursion.py
Farnazmdi: I agree, and they appear in the denominator, so never should be zero.

lineage/tests/test_BaumWelch.py
Farnazmdi: I remember that it was set to be -inf at first, so it improves over time, right @shak360?

shak360: Yes, the initial condition is just -inf because that makes sure that it will always pass the first check that that the next step will always have a higher likelihood than the 0th step.

aarmey: But if you comment this out, the LL after running fitting for two steps is still -inf.

lineage/UpwardRecursion.py
shak360: Yes they're being returned as `nans` and cannot be compared to 0.0. This implies that the problem is in `MSD`, the marginal state distribution.

lineage/tests/test_BaumWelch.py
shak360: Just fixed these. Getting the `NF` for the whole tree requires running three functions (`get_leaf_NFs`, `get_leaf_betas`, and `get_nonleaf_NFs_and_betas`) not just `get_leaf_NFs`.

shak360: Likelihoods are operating as expected (only ever increase through each iteration). Now the error is back in the aforementioned ` assert NF_array[leaf_cell_idx] > 0.0`. Will keep looking, but issue seems to be with `MSD`.

shak360:
I've isolated the error to `EL`. I was wrong about my previous assumption that `assert NF_array[leaf_cell_idx] > 0.0` breaking was due to `MSD`. The `nan` in the product `NF_array[leaf_cell_idx] = np.sum(MSD_array[leaf_cell_idx, :] * EL_array[leaf_cell_idx, :])` is actually coming from `EL`. Here is one of the `EL`s printed below:
image

aarmey:
Can we get this to a point where the builds pass and merge? Worried this has strayed really far from master. Then figure changes can continue on a new branch.

shak360:
Yes, give me 1-2 hours, it'll be ready to merge.

shak360:
Just linting left to do.

PR #166, opened 2019-09-15
Build with manubot

aarmey:
Allows annotation of built manuscript. Simpler citations.


aarmey:
Only issue is it's getting stuck on an equation somewhere in the methods.

shak360:
I tried finding the \LaTeX error but can't really get anywhere. Does this error `jinja2.exceptions.TemplateSyntaxError: unexpected char '\\' at 29050` ring any bells for you @aarmey

shak360:
I have a feeling it might be due to the use of `\bm` in the `\LaTeX` writing since that is a package that needs to be included if it needs to work.

aarmey:
I was thinking the same but want to try deleting just the first or last half of the file when I get a chance to ensure it's not something more specific...

aarmey:
@shak360 error I was seeing was from the equation you deleted from the methods. It looks like you committed a bunch of figure files here. Can you clean this up?

aarmey:
@shak360 in fact, please revert your commits here, then branch off this and work on a separate branch. `\bm` is not the problem. The jinja errors are before latex is called, so there's a syntax error somewhere when you see them.

shak360:
Ready when you are @aarmey Was the error with the display-style LaTeX equation?

aarmey:
Thanks. There were some extraneous curly braces. They were correct syntax but didn’t work with ninja for some reason. Removing them didn’t affect the output, so it was an easy fix.

PR #167, opened 2019-09-18
Phase emission

Farnazmdi:
I added another file `StateDistribution2.py` to separate G1 and G2 emissions and made setup functions for making figure6 based on the new distribution function. Tests pass.


Farnazmdi:
Seems it is failing because of something in the manuscript

Farnazmdi:
image

PR #172, opened 2019-09-29
Added the pruning based on time

shak360:
Lineages are now created by default with a number of cells equal to (2**11)-1 which is 2047 cells. This enables us to create lineages long past the experiment run time, which is now a input argument for lineage creation. That is to say, all unpruned lineages are always going to be 2047 cells long. Pruned lineages then can be pruned in 1 of 3 ways (or just not pruned at all):

**if `fate` is chosen: daughter subtrees of cells that die are excised.
if `time` is chosen: daughter subtrees of cells with `endT` greater than some desired experimental time are excised.
if `both` is chosen: both of the above prune rules are applied (`OR` statement).**

A `Time` class is now added which each cell now contains under `cell.time` which contains the `startT`, `endT`, and `lifetime`. It seemed appropriate to put the definition of the time related functions into the `StateDistribution` file. The use of the new functions (not the definition of) are in the `LineageTree` file. The `prune_boolean` and the `prune_condition` statement are now cleaner and easier to use.

Functions that dealt with time are now rewritten to accommodate this.

Old tests were removed and new tests were added.

Figure code has NOT yet been changed and WILL break. Will change in a later branch.

### DIFF
--- a/lineage/LineageTree.py
+++ b/lineage/LineageTree.py
@@ -3,7 +3,7 @@ import scipy.stats as sp
 from copy import deepcopy
 
 from .CellVar import CellVar
-from .StateDistribution import die_prune_rule
+from .StateDistribution import assign_times, die_prune_rule
 
 
 # temporary style guide:
@@ -99,12 +99,14 @@ class LineageTree:
             self.output_list_of_gens = self.pruned_list_of_gens
             self.output_leaves_idx = self.pruned_leaves_idx
             self.output_leaves = self.pruned_leaves
+            assign_times(self)
         else:
             self.output_lineage = self.full_lin_list
             self.output_max_gen = self.full_max_gen
             self.output_list_of_gens = self.full_list_of_gens
             self.output_leaves_idx = self.full_leaves_idx
             self.output_leaves = self.full_leaves
+            assign_times(self)
 
     def __len__(self, prune_boolean):
         if prune_boolean:
--- a/lineage/LineageTree.py
+++ b/lineage/LineageTree.py
@@ -23,7 +23,7 @@ class LineageStateStats:
 
 
 class LineageTree:
-    def __init__(self, pi, T, E, desired_num_cells, desired_experiment_time, prune_condition='both'):
+    def __init__(self, pi, T, E, desired_experiment_time, prune_condition='both'):
         """
         A class for the structure of the lineage tree. Every lineage from this class is a binary tree built based on initial probabilities and transition probabilities given by the user that builds up the states based off of these until it reaches the desired number of cells in the tree, and then stops. Given the desired distributions for emission, the object will have the "E" a list of state distribution objects assigned to them.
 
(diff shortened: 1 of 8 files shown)


lineage/LineageTree.py
```diff
@@ -140,23 +143,42 @@ def _generate_lineage_list(self):
                 self.full_lin_list.append(left_cell)
                 self.full_lin_list.append(right_cell)
 
-            if len(self.full_lin_list) >= self.desired_num_cells:
+            if len(self.full_lin_list) >= 2**11 - 1:
```
shak360: Unpruned lineages by default are now 2047 cells long.

lineage/LineageTree.py
```diff
@@ -140,23 +143,42 @@ def _generate_lineage_list(self):
...
     def _prune_lineage(self):
-        """ This function removes those cells that are intended to be remove from the full binary tree based on emissions.
-        It takes in LineageTree object, walks through all the cells in the full binary tree, applies the pruning to each cell that is supposed to be removed, and returns the pruned list of cells.
+        """ This function removes those cells that are intended to be remove
+        from the full binary tree based on emissions.
+        It takes in LineageTree object, walks through all the cells in the full binary tree,
+        applies the pruning to each cell that is supposed to be removed,
+        and returns the pruned list of cells.
         """
+        assign_times(self)
```
shak360: Gives each cell the `time` attribute.

lineage/LineageTree.py
```diff
@@ -140,23 +143,42 @@ def _generate_lineage_list(self):
...
         self.pruned_lin_list = deepcopy(self.full_lin_list)
         for cell in self.pruned_lin_list:
-            if prune_rule(cell):
-                _, _, self.pruned_lin_list = find_two_subtrees(
-                    cell, self.pruned_lin_list)
-                cell.left = None
-                cell.right = None
-                assert cell._isLeaf()
+            if self.prune_condition == 'both':
```
shak360: User-decided pruning.

lineage/StateDistribution.py
```diff
@@ -73,49 +73,98 @@ def __repr__(self):
...
                              10 * (np.random.uniform()),
                              0,
                              1)
 
-# Because parameter estimation requires that estimators be written or imported, the user should be able to provide
-# estimators that can solve for the parameters that describe the distributions. We provide some estimators below as an example.
-# Their use in the StateDistribution class is shown in the estimator class method. User must take care to define estimators that
-# can handle the case where the list of observations is empty.
 
+class Time:
```
shak360: `Time` class and the associated functions.

shak360:
Again figures will NOT build, will fix on a later branch. Need to rearrange based on comments from last meeting.

shak360:
@Farnazmdi do you mind approving this?

shak360:
Thanks all

PR #175, opened 2019-10-25
added a new assert statement

jclagarde:


shak360:
Hey @jclagarde. Good work today. I think you'll definitely learn a lot about tHMM and how to use Git and Python by adding asserts and tests to these files. 

Like we talked about, feel free to add any asserts or tests that you think should pass (which you can test with `pytest -k test_CellVar.py` or any of the other files you're editing). Feel free to start with test_CellVar.py then either `test_StateDistribution.py` or `test_LineageTree.py`. Hopefully, doing so will get you comfortable with creating your own lineage trees in the future, and ultimately running our model on them (wherein you'll be able to write asserts and tests for `test_tHMM.py` and test_BaumWelch.py`).

Another thing you can do is write comments on existing tests or asserts if it helps you understand what is going on.

===

For starters, here's an assignment (feel free to do things not on this list, but being able to do the things on this list is sufficient to move on, in my opinion, to `test_StateDistribution.py` or `test_LineageTree.py`) (Also, I'm not expecting a write-up or anything, most of these questions are just to guide your thinking):

Easy:
[] add explanatory but brief comments (<6 words) to the lines in the function at https://github.com/meyer-lab/lineage-growth/blob/f7e19bae42378820696b91fa4e3393e23b208659/lineage/tests/test_CellVar.py#L75

Medium:
[] add asserts regarding which cells should and should not be leaves to the function at https://github.com/meyer-lab/lineage-growth/blob/f7e19bae42378820696b91fa4e3393e23b208659/lineage/tests/test_CellVar.py#L109 (hint: mimic some of the asserts in the function right below it)

[] likewise add asserts regarding which cells should and should not be root parents to the function at https://github.com/meyer-lab/lineage-growth/blob/f7e19bae42378820696b91fa4e3393e23b208659/lineage/tests/test_CellVar.py#L126 (hint: mimic some of the asserts in the function right above it)

Hard:
[] test what happens when the probability of switching from state 0 to state 1 is 100% and the probability of switching from state 1 to state 0 is 100% (what will the transition matrix look like? is this a valid transition matrix?  what should the tree of cells look like if we started at state 0? what about starting at state 1?). Your test for this should be similar to the function at https://github.com/meyer-lab/lineage-growth/blob/f7e19bae42378820696b91fa4e3393e23b208659/lineage/tests/test_CellVar.py#L176 . Don't forget to add docstrings (`"""message under function"""`) and comments to your code.

[] test what happens when the probability of switching from state 0 to state 1 is 100% and the probability of switching from state 1 to state 0 is 0%. Try to ask yourself the same questions as above, and model your resulting test as the function linked above.

[] write a test that utilizes the transition processes shown here (ask what would the transition matrix look like?)

image

This is from the paper, if you're curious (https://www.sciencedirect.com/science/article/pii/S0092867411008245). What are some things you can expect from transitions like these? Can you write asserts in the test that can verify your predictions? Why do you think all the other transition matrices are all identity matrices in the other tests (what is difficult about testing transition matrices that aren't identity matrices?)

aarmey:
@shak360 @jclagarde is this something we should ultimately merge?

shak360:
No it isn't; this was just for showing JC some of the repository. I thought I closed this PR though...

aarmey:
I opened it to update it. Will close now.

PR #179, opened 2019-10-31
adding stationary distribution code

shak360:
Calculates the stationary distribution given a transition matrix.


aarmey:
This is the right stationary distribution if our phenotype of interest isn't growth. However, if the phenotype influences the relative abundance of each subpopulation, then we have to account for that. For example, if you have the same transition matrix, but state A grows 10X faster, that should change the abundance of A. I think it's probably something as simple as scaling each population by its relative growth rate, but I'm not sure.

shak360:
Linted. Should be ready for review/merge @aarmey .

PR #180, opened 2019-11-01
Figure 3

shak360:
image

Realized I calculated the number of bins incorrectly in the earlier versions of these histograms... I was multiplying the delta time by the total experimental time, when I should have been dividing. Fixed the error, recreated the figure and added a sister chart showing the distribution and is now one of the figures that is auto-generated.


shak360:
Keep in mind orange cells live 10 times longer than blue cells in this model on average. 

orange: gamma a is 20 and scale is 5 which means a mean of 100 hours
blue : gamma a is 10 and scale is 1 which means a mean of 10 hours

shak360:
Combined the figures
image

shak360:
Linted. Should be ready for review/merge @aarmey.

PR #181, opened 2019-11-03
Figures 1 & 2

shak360:
Sorry I just needed to generate these because I've been worried that the underlying way to synthesize cells was wrong... but these figures show that they're valid.

Figure 1:
image
Key take away: the population distribution tends to the stationary distribution of the Markov transition matrix (phew...)

Figure 2:
image
Key take away: the population distribution does not tend to the stationary distribution when the lineages are pruned


shak360:
Guh I think my work on Figure 3 in the other branch snuck its way into this branch... It should be fine...

shak360:
Combined the figures
image

shak360:
Pretty sure this is ready for a review/merge. Getting this error from Jenkins, however,
```
+ tlmgr install type1cm
===============================================================================
tlmgr itself needs to be updated.
Please do this via either
  tlmgr update --self
or by getting the latest updater for Unix-ish systems:
  http://mirror.ctan.org/systems/texlive/tlnet/update-tlmgr-latest.sh
and/or Windows systems:
  http://mirror.ctan.org/systems/texlive/tlnet/update-tlmgr-latest.exe
Then continue with other updates as usual.
===============================================================================
tlmgr: Terminating; please see warning above!
tlmgr: package repository http://ctan.mirrors.hoobly.com/systems/texlive/tlnet (verified)
```

PR #182, opened 2019-11-03
adding try catch to fitting

shak360:
Added the small try catch statement to catch exceptions from `NaN`s. Would like to treat this in a better (rigourous, mathematical) fashion but this works really well for now.

### DIFF
--- a/lineage/Analyze.py
+++ b/lineage/Analyze.py
@@ -25,18 +25,15 @@ def Analyze(X, numStates):
     LL {}:
     """
 
-    run = True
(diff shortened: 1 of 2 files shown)


shak360:
Looks like it never actually takes more than 2 tries (most of the time it fits with just one try).

shak360:
Looks great, thanks Dr. Meyer @aarmey. Will merge and delete the branch. The resulting change should make all other PRs build (apart from linting; the quality gate is at 91).

PR #186, opened 2019-11-05
Kl

Farnazmdi:
not ready yet.


shak360:
Fix a couple of linting errors manually, and the build should start passing @Farnazmdi

Farnazmdi:
image
Is some figure like this good to include? As a complement to the KL itself. @aarmey  @shak360 
image

Farnazmdi:
Oops, the order is vice versa.

shak360:
Are those the gamma distributions for each of the points?

Farnazmdi:
yes, those in the same color are the ones that we are taking the KL divergence for them, first, they are very far apart (blue ones) and then they get closer (gray).

shak360:
Good work 

Hmmm I'm not sure what Dr. Meyer will say, but I say it's better to have more points on the KL-divergence plot, just to see more of the trend (maybe try using `np.linspace` or `np.logspace` to generate parameters). I feel like the the plot with the gamma distributions can be explained verbally through a figure caption or a paragraph but I'm not sure. It does, however, visually explain what's going on though...

aarmey:
Maybe the distribution plot would work better as a split violin?

https://seaborn.pydata.org/examples/grouped_violinplots.html

Everything on top of one another is busy. I think the plot does have value in showing what we're doing.

aarmey:
And yes, a few more points on the accuracy graph could be helpful.

lineage/Analyze.py
```diff
@@ -225,3 +226,10 @@ def accuracyG(tHMMobj, all_states):
         acuracy_holder.append(100 * acc)
 
     return acuracy_holder
+
+
+def kl_divergence(p, q):
```
aarmey: Why not use [`scipy.stats.entropy`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.entropy.html)?

lineage/figures/figure12.py
```diff
@@ -0,0 +1,120 @@
...
+    T = np.array([[0.50, 0.50],
+                  [0.50, 0.50]])
+
+    state0 = 0
+    state1 = 1
+    gamma_loc = 0
+    bern_p0 = 0.99
+    bern_p1 = 0.88
+    a0 = np.linspace(5.0, 17.0, 10)
+#     a0 = [5.0, 10.0, 15.0, 12.0]
```
aarmey: Remove commented out lines if you're not using them.

Farnazmdi:
How about this?
image

shak360:
I think the distributions should be on different sides of the violins, not on the same side. But this is good I think.

Farnazmdi:
You're right, I searched to plot it that way, but couldn't find it. I need a "hue" that decides what should be on each side of the distribution. I couldn't think of any simple way.

Farnazmdi:
image
The x-axis is off but the results should be correct. Does it make sense?

aarmey:
I can figure out how to get the split plot once the code is in the figure file.

Note that the changes should be removed from the ipynb. It's supposed to be a guide.

aarmey:
> image
> The x-axis is off but the results should be correct. Does it make sense?

Is there very much run-to-run variation? Because this is plotted differently from all the other accuracy plots. We should have a consistent style.

Farnazmdi:
@aarmey Thank you. Sure, I just wanted to keep it somewhere, I will move it to the figure12.py file.

Farnazmdi:
> > image
> > The x-axis is off but the results should be correct. Does it make sense?
> 
> Is there very much run-to-run variation? Because this is plotted differently from all the other accuracy plots. We should have a consistent style.

No, I will remove the solid line, I am using the same style as other figures.

Farnazmdi:
image

I am not sure what is the problem with the xticks.

Farnazmdi:
image

aarmey:
Oh perfect!

shak360:
Linting should make this PR mergable @Farnazmdi

Farnazmdi:
will do later today @shak360

lineage/figures/figure12.py
```diff
@@ -0,0 +1,213 @@
...
+                                         size=500),
+                            sp.gamma.rvs(a=a1[8],
+                                         loc=gamma_loc, scale=scale1[8],
+                                         size=500),
+                            sp.gamma.rvs(a=a0[7], loc=gamma_loc,
+                                         scale=scale0[7], size=500),
+                            sp.gamma.rvs(a=a1[7], loc=gamma_loc,
+                                         scale=scale1[7], size=500),
+                            sp.gamma.rvs(a=a0[6], loc=gamma_loc,
+                                         scale=scale0[6], size=500),
```
aarmey: Surely there's a loop you can use to do this?

lineage/figures/figure12.py
```diff
@@ -0,0 +1,213 @@
...
+                                         scale=scale1[1], size=500),
+                            sp.gamma.rvs(a=a0[0], loc=gamma_loc,
+                                         scale=scale0[0], size=500),
+                            sp.gamma.rvs(a=a1[0], loc=gamma_loc,
+                                         scale=scale1[0], size=500)))
+
+    dists = pd.DataFrame(columns=['lifetime [hr]', 'distributions', 'hues'])
+    dists['lifetime [hr]'] = dist1
+    dists['distributions'] = 1000 * (['d1']) + 1000 * (['d2']) + \
+        1000 * (['d3']) + 1000 * (['d4']) + 1000 * (['d5']) + \
```
aarmey: You can loop and append these to the DataFrame in each iteration. I think that'd be cleaner.

lineage/Analyze.py
aarmey: Still wondering about this.

lineage/figures/figure12.py
Farnazmdi: Oh, that's right. will do right away.
And I'm not using the function, I'm using the entropy, just forgot to remove it. Will do that, too.
Thanks.

Farnazmdi:
I couldn't figure out the reason for the instability of the build @shak360, has it happened before?

aarmey:
@Farnazmdi I think this is related to the power going out last night. Retry the builds and I think it should work.

Farnazmdi:
I've built it 2 times since this morning @aarmey 
Seems the problem is here:
image

aarmey:
This is the most recent error:

```
=================================== FAILURES ===================================
_____________________ TestModel.test_exponential_estimator _____________________

self = <lineage.tests.test_StateDistribution.TestModel testMethod=test_exponential_estimator>

    def test_exponential_estimator(self):
        """
        Testing the exponential estimator,
        by comparing the result of the estimator
        to the result of scipy random variable generator.
        """
        exp_obs = sp.expon.rvs(
            scale=50, size=1000)  # exponential observations
        self.assertTrue(45 <= exponential_estimator(
>           exp_obs) <= 55)  # +/- 5 of beta
E       AssertionError: False is not true

lineage/tests/test_StateDistribution.py:159: AssertionError
```

Farnazmdi:
I guess if we add another file, then linting would be hard! Could we set the threshold for linting a little higher? @aarmey

aarmey:
I still see plenty of easily fixed and reasonable linting warnings. :-)

Farnazmdi:
Oh, sorry, the ones I saw were for having too many arguments, or long functions? I probably haven't seen all of them! @aarmey

PR #194, opened 2019-12-02
merging figures 2+3, figure cleanup

shak360:
1. deleted figures 4, 5, and 6 
reasoning: these figures described the results of analyzing and fitting unpruned lineages

2. merged figures 2 and 3 into figure 2. removed the aspects of the figures that involved fitting unpruned lineages. Figures 1, 3, 4, 5, and 6 are now blank. 

2.5 set an x-axis limit of the experimental time in figure 2 when tracking cells over time. the x-axis for the cells tracked over generations is now integral as opposed to floaty. Generations are also now 1-indexed.

Result:
image

I imagine tweaking the state parameters for this figure will give us a good idea for what parameters to choose when creating lineages that look like cell experiments (prior to conducting cell experiments).

3. more work on the Guide to tHMM iPython notebook for JC's final assignment and for an end-user.


shak360:
need to lint

PR #197, opened 2019-12-02
picking the best likelihood

shak360:
Analyze now picks the result of the best likelihood off of 5 tries. Figures are changed to accommodate this change (no longer run until an accuracy > 50% is chosen.)


shak360:
image
 

This is pretty good all things considered. I still have to clean this figure up but it'll be on another branch I think. Not sure if you wanna move forward with showing AIC.

PR #206, opened 2019-12-11
Multi p

Farnazmdi:
Still need to figure out: It just generates the first figure correctly and doesn't go through the other ones. 

Changes made in this PR: 
1. I changed the figure names, so it just deals with those files that they generate a figure (not empty).
2. Changed `genFigures.py`
3. Using `Pool` from `multiprocessing`.


aarmey:
@Farnazmdi what are you trying to do here? Generate each figure in parallel? Make can already do that.

Farnazmdi:
Yes, I wanted to run them at the same time. Huh, I didn't know that!
So when we talk about multiprocessing, which part do we want to run in parallel?

aarmey:
Analyze() fits the model to a new set of random data X number of times. The issue is about running each of those trials (within a single plot) in parallel.

aarmey:
It would help to write this out in the issue, because I've made this correction more than once (to Shakthi, not to you).

Farnazmdi:
Thanks, I will close this one and open another PR for that.

Issue #216, opened 2020-01-19
Fix accuracy function to account for more than 2 states and not use true labels


shak360:
Different states are correctly identified but their order may not necessarily be the same as the order of the states originally used to define the lineage. This makes it difficult to calculate the accuracy of state assignments. When two states are used, we've implemented it where it calculates a "distance" (not really a distance) between the parameters of the estimated state and the true state and then either maintains (when the distance between estimated state 1 and true state 1 is small) or flips (when the distance between estimated state 1 and true state 1 is large) the estimated state label. The problem with this approach is that it uses the true labels which the model should be ultimately blind to if making a decision to switch label names. It also only really works well for two states.

Ideally, and as discussed before, we should be using something like this:

https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mutual_info_score.html#sklearn.metrics.mutual_info_score

https://scikit-learn.org/stable/modules/generated/sklearn.metrics.adjusted_mutual_info_score.html#sklearn.metrics.adjusted_mutual_info_score

Note that for flipping the estimated state labels (or in the case of more than 2 state labels, assigning different state labels based on the true state labels), there are three things that need to be changed:

1. the individual elements of the transition matrix need to be moved (we're currently doing this incorrectly... which is why our transition matrix accuracy plot looks _terrible_ and has two asymptotes instead of just one asymptote toward 0 error.) Currently, we just take the transpose if state 0 needs to become state 1 and vice versa. However, taking the transpose only fixes the values on the off-diagonal (01 -> 10 and 10 -> 01). Taking the transpose doesn't move the elements at 00 and 11 which actually need to be moved to 11 and 00 respectively. Furthermore, a transpose will not work for something like 3 states where state 2 needs to become 3, state 1 needs to become state 2, and state 3 needs to become state 1. Ideally, a "map" of some kind needs to made which maps elements' indices to new indices, and a new transition matrix needs to be filled out. I'm having trouble thinking of how one would make a map of how to move the state labels _without_ using the true labels and their underlying states and cells. I think that since it is for the purpose of displaying accuracy and not for analysis, it should be fine apart from mutual information score.

2. The same problem holds for the initial probability vector, but is much easier to handle because it's a single-dimensional vector. Solving the "map" aspect of the transition matrix lets you flip the elements of this vector easily.

3. Again creating the map will let you change the elements of the Emissions list.

The map function would need to do two things:

[...]

PR #217, opened 2020-01-19
Sv final hw

shak360:
Final practice for JC. Contains samples on how to analyze lineages, and some questions to test understanding. Let me know if you want to add more or less to this practice @aarmey. I avoided adding a question on creating three state lineages because of some issues we have with calculating accuracy (see #216). I figure that this assignment can also double as the guide to the code. If needed, a second guide can be written to talk about how users can write `StateDistribution` files and how to create three or more state lineages. Once approved, I can let JC work on it.

See note book here:
https://github.com/meyer-lab/lineage-growth/blob/sv-finalHW/Guide_to_tHMM.ipynb


shak360:
Failures due to changes to `accuracy`.

PR #220, opened 2020-01-24
Sv fix accuracy

shak360:
Removed previous accuracy function. Wrote new `results` function in its place. Computes more metrics fairily (without permutation) and can handle an arbitrary number of states. Also permutes the states after calculating metrics.

See #216 

Explanation:
The `switcher_map` is a map between predicted states before and after switching by using the (index, value) pairs in the list `switcher_map`. For example, the following list `[0 1 2 3]` implies that the states don't need switching because the index and the values match. Something like `[1 0 3 2]` indicates that the predicted labels for `0` need to be switched to `1`,  `2` to `3` and so on. The map was calculated based on the KL_divergence of the parameterized distributions of each label. The transition matrix, pi vector, and emissions list are then mapped to different labels using the `switcher_map`. 

TODO:
1. Handle the case where the `switcher_map` has non-unique duplicate values.
~~2. Fix the figures to handle this new function to prep for parallelizing figures.~~
3. Write a new advanced guide/assignment that deals with fitting lineages with >2 states.


shak360:
This branch is ready for merging @aarmey, if you want to check and approve or branch off of this and add your changes.

lineage/Analyze.py
```diff
@@ -169,7 +266,7 @@ def get_stationary_distribution(transition_matrix):
     c = np.zeros((K + 1, 1))  # 9
     c[K, 0] = 1
     p = np.linalg.solve(BT_B, np.matmul(B.T, c)).T  # 11
-    assert np.allclose(np.matmul(transition_matrix.T, p.T), p.T)  # 12
+    # assert np.allclose(np.matmul(transition_matrix.T, p.T), p.T)  # 12
```
aarmey: Why comment this out?

shak360: I was receiving some error about how `np.matmul` didn't like the transition matrix because it was nonsingular and I lazily commented it out.

Issue #211, opened 2020-01-04
Have pdf calculated through a fused, numba-compiled function


aarmey:
Fitting spends most of its time calculating the pdf:

https://github.com/meyer-lab/lineage-growth/blob/cfad4a2951a4148c4cc863f325ca7008e1d08cfc/lineage/StateDistribution.py#L27

The scipy functions broadcast over observations in a simple loop. This can probably be much more efficient if we just calculate it ourselves.

shak360:
See #228 Done.

PR #228, opened 2020-02-19
Adding numba decorators 

shak360:
JC and I have written functions (using only `numpy` and python's STL `math`) for the Bernoulli and the gamma probability distribution functions. Numba JIT decorators were then added.


shak360:
See #211

PR #245, opened 2020-03-02
Get rid of one more loop.

aarmey:
Should take out one more slow spot in the first profiling output.


aarmey:
Check that the build output matches master before merging.

PR #249, opened 2020-03-04
Vector viterbi

aarmey:
I've tried to vectorize across state_j, but broke something. Could someone take a look?


aarmey:
Will take some work to fix this.

aarmey:
Should be fixed...

Issue #230, opened 2020-02-24
model performance vs relative heterogeneity 


shak360:
Hi @jclagarde ,

We'll try to do this next assignment in two to three sub-assignments (I'd also like for you to push the changes to the Guide, if you haven't done so already). Recall that the main goal is to add another figure to the project. 

**Figure 5. Model performance vs the relative amount of heterogeneity**

Things you should know before you begin:
You can look at the figures generated by our project in one of two ways:
1. Access our CI client Jenkins. CI stands for continuous integration and is a software tool designed for managing large codebases or code that is continuously changing. You can access it here: 

https://transduc.seas.ucla.edu/job/meyer-lab/job/lineage-growth/job/master/

Previous "builds" or versions of our projects (and other projects from Dr. Meyer's lab) can be found here. Feel free to play around. Do not click on anything that looks dangerous. You can see the most updated "builds" or versions of our manuscript/figures at the top of that page. You can also see how various branches of our project can affect these builds when you examine those branches on Jenkins. I can show you how to navigate this website more in person.

2. On your own Aretha server account. First, run through your basic git commands:
```
git status
git checkout master
git pull
git branch jc-fig
git checkout jc-fig
```
Go to your lineage folder and then up one folder (that is, you should see a file called `makefile`, the Guide that you've been working on, and the `lineage` folder itself). Check if there exists a folder called `output`. If not, open up terminal, `cd` into this `lineage-growth` directory (that is, if you `ls -ltr`, you should see the `makefile`, the guide, and the `lineage` folder) or whatever you have it named as, and then type in `make`. What this does is, it runs the commands listed in the makefile. It does many things that are relevant for users, but for now, we can focus on how it makes figures. Once `make` is done running, run `cat -n makefile` in the terminal; this displays the content of the makefile, which is just a textfile of shell commands, onto the terminal. You can also open the `makefile` in the text editor by double-clicking on the file on the left file explorer bar. 

Tasks:

0. Because we're going to add another figure, let's add the number `7` to line 7 in the makefile, to the variable `flist`. Save the file and exit. Recall your git commands for adding, commiting, and pushing:
```
git add -A
git commit -m "adding 7 to makefile"
git push
```
Recall that the the first `push` on a new branch will require you to set the origin. Follow what git prompts you to do.

[...]

PR #261, opened 2020-03-12
Figure 5

jclagarde:
final part of issue #230


lineage/figures/figure5.py
```diff
@@ -1,14 +1,67 @@
...
+def accuracy_increased_cells():
+    """
+    Calculates accuracy and transition rate estimation over an increasing number of cells in a lineage for an pruned two-state model.
+    """
+
+    # Creating a list of populations to analyze over
+    list_of_populations = []
+    
+    # Create a list of transition matrices to transition over called list_of_Ts
+    list_of_Ts = []
```
shak360: You need to call your `makeTs` function and store the result in `list_of_Ts`

lineage/figures/figure5.py
```diff
@@ -1,14 +1,67 @@
...
+    
+    # Iterate through the transition matrices
+    for T in list_of_Ts:
+        # Creating an unpruned and pruned lineage
+        lineage = LineageTree(piiii, T, E, (2**12) - 1, experiment_time, prune_condition='both', prune_boolean=True)
```
shak360:
```suggestion
        lineage = LineageTree(piiii, T, E, (2**12) - 1, experiment_time=1e9, prune_condition='both', prune_boolean=True)
```

lineage/figures/figure5.py
```diff
@@ -1,14 +1,67 @@
...
+
+        while len(lineage.output_lineage) < 16:
+            del lineage
+            lineage = LineageTree(piiii, T, E, (2**12) - 1, experiment_time, prune_condition='both', prune_boolean=True)
```
shak360:
```suggestion
            lineage = LineageTree(piiii, T, E, (2**12) - 1, experiment_time=1e9, prune_condition='both', prune_boolean=True)
```

shak360:
image

@aarmey Figure 5 is done.

shak360:
I'll make an even more common figure4Common function in a new branch and change figure 6 with it. This PR will is also the last of JC's assignments in #230

aarmey:
The code all looks good. Any ideas about what's happening with the transition matrix estimation?

shak360:
Hmmmm, The Initial probability matrix estimation I'm not worried about because these are all single lineages so it doesn't make sense to even plot those.

In a similar scenario:
image

(with normal transitions instead of transitions going from 0.5 to 0 or 0.5 to 1) this is what we get

Issue #227, opened 2020-02-04
Profile again


shak360:

shak360:
![image](https://user-images.githubusercontent.com/20716101/73793713-cb804a80-475b-11ea-98cb-f32a458375e3.png)

shak360:
![image](https://user-images.githubusercontent.com/20716101/74979040-a853cd00-53e2-11ea-8ca9-fec3a9d14451.png)

shak360:
Looks like the hold is up in pruning lineages, which isn't utilized in the model anyway, only in lineage synthesis. This is to be expected because this is done recrusively through the tree and walks through the tree several times.

shak360:
It was suggested that we individually profile the figure creation process (not the testing) as I usually do. We can identify which figure takes the most time. My bet is on figure 5 (AIC).

shak360:
To profile:
```bash
python3 -m cProfile -o profile genFigures.py 5
gprof2dot -f pstats --node-thres=5.0 profile | dot -Tsvg -o profile.svg
```

shak360:
![image](https://user-images.githubusercontent.com/20716101/75200825-a314bc00-571b-11ea-8068-b6aebe4d03f7.png)
AIC figure

shak360:
![image](https://user-images.githubusercontent.com/20716101/75201207-b5dbc080-571c-11ea-814a-73b2629197da.png)
Figure 4 (analyzing over an increasing number of lineages or starting cells)

shak360:
Seems like making pruning faster would be the best bet in terms of speeding up figures.

shak360:
Wrote a short script to generate all figures at once and then profiled it:
![image](https://user-images.githubusercontent.com/20716101/75288951-301a4c80-57d2-11ea-9f77-2024040f804e.png)

shak360:
![image](https://user-images.githubusercontent.com/20716101/75290253-d7987e80-57d4-11ea-8c82-e6e824a21f50.png)
Figure 6 (performance vs distribution divergence)

aarmey:
@shak360 note that you have to remove any multithreading to get accurate profiling. `thread.lock` just means the thread is waiting for another to finish.

shak360:
![image](https://user-images.githubusercontent.com/20716101/75597293-93fa7a80-5a49-11ea-87b6-5bcb951668e3.png)

shak360:
[...]

PR #264, opened 2020-03-12
Lineage pruning REFACTOR

shak360:


shak360:
need to lint

Farnazmdi:
Figure 2 looks weird in terms of dimensions.
image

Farnazmdi:
This one, too.
image

shak360:
Sorry, I know we discussed not having gargantuan pull requests but this sort of change necessitated it I think. This removes the `deepcopy` in lineage pruning saving both space and time in lineage creation. Furthemore, `**kwargs` are now used in both cell and lineage instantiation, to avoid extremely verbose calls to create them. Now, cells don't always have by default a `left` and `right` daughter linkage to `Nonetype` objects. I also removed the uneccesary underscores in function names and made sure that the variable `num_states` is called `num_states` across the code (instead of sometimes `numStates`). Cells can also now be leaves in one of two ways: being terminal (i.e., being the cells at the very end of the tree (I call these terminal leaves) and being cells whose daughters are both censored. Note that pruned is no longer used anywhere in the code, censored made more sense to me. Lineages now utilize a censor condition which can be 0,1,2, or 3. This made more sense than using string arguments, and I think makes it easier if someone wanted to add more or different ones. By default, the censor condition is 0.

shak360:
I can focus on the figures in a later branch

shak360:
`sv-pruneSpeed`
figure1 is done after 4.444711685180664 seconds.
figure2 is done after 142.2366909980774 seconds.
figure3 is done after 180.15815329551697 seconds.
figure4 is done after 189.2325563430786 seconds.
**figure5 is done after 60.49672508239746 seconds.**
figure6 is done after 146.81397128105164 seconds.
figure7 is done after 85.04389572143555 seconds.
===
`master`
figure1 is done after 2.576735734939575 seconds.
figure2 is done after 39.26531362533569 seconds.
figure3 is done after 86.82021522521973 seconds.
figure4 is done after 103.25810480117798 seconds.
figure5 is done after 80.15113925933838 seconds.
figure6 is done after 90.22808051109314 seconds.
figure7 is done after 45.43596792221069 seconds.

Not sure if this is any faster or not, I'll try comparing the profiles between my branch and master

shak360:
@aarmey The new profile results are in #227, feel free to go into the code to change anything you think should be faster. I wanna say that the code is more readable now, and possibly easier for you to make changes to.

shak360:
In figure 5, I zoomed in on the region we were talking about yesterday. Figure 4 looks the same as master for me:
master:

image

current branch:

image

lineage/LineageTree.py
```diff
@@ -44,234 +27,137 @@ def __init__(self, pi, T, E, desired_num_cells, desired_experiment_time, prune_c
...
+        applies the pruning to each cell that is supposed to be removed,
+        and returns the censord list of cells.
+        """
+        self.output_lineage = []
+        for idx, cell in enumerate(self.full_lineage):
+            if self.censor_condition == 0:
+                self.output_lineage = self.full_lineage
+                break
+            elif self.censor_condition == 1:
+                if fate_censor_rule(cell):
```
aarmey: Rather than find and copy the subtrees over and over, why not have a method `censor_cell_and_daughters()`, then have it be a recursive function that walks down the tree until a cell is a leaf?

shak360: Yea that would avoid the huge number of for loops, you're right

shak360: We most likely have the recursions written as functions already; I'll try to see if I can implement this using those. I can also do this on a separate branch, or on this one if you'd like.

aarmey: We can merge then make this change.

PR #270, opened 2020-03-14
add likelihood to aic fig

Farnazmdi:


lineage/figures/figure7.py
```diff
@@ -56,7 +56,10 @@ def run_AIC(Trate, E, num_to_evaluate=10):
         # Analyze the lineages in the list of populations
         output = run_Analyze_over(list_of_populations, num_states_to_evaluate)
         # Collecting the results of analyzing the lineages
-        for idx, (tHMMobj, _, LL) in enumerate(output):
+        for idx, (tHMMobj,pred_states_by_lineage,_) in enumerate(output):
+            # Get the likelihood of states
+            LLtemp = LLFunc(T, pi, tHMMobj, pred_states_by_lineage)
+            LL = np.sum(LLtemp)
```
aarmey: This sum is over trees right?

Farnazmdi: Yes, It is summing over the trees. They have more than 1 lineage in each population.

PR #273, opened 2020-03-17
Sv prune speed2

shak360:


shak360:
@aarmey Is it even worth comparing it like this (see below)? I _know_ my branch HAS to be faster, the added `if` statement should avoid a ton of `for` loops being calculated. It has to be faster, but these completion times aren't reflecting it. Figures 3 and 4 should at least change.

My branch:

- figure1 is done after 4.0449440479278564 seconds.
- figure2 is done after 84.71237874031067 seconds.
- figure3 is done after 79.11824250221252 seconds.
- **figure4 is done after 78.48520708084106 seconds.**
- figure5 is done after 66.78407788276672 seconds.
- figure6 is done after 95.9872305393219 seconds.
- figure7 is done after 60.79370474815369 seconds.

master:

- figure1 is done after 2.765455961227417 seconds.
- figure2 is done after 63.29394769668579 seconds.
- figure3 is done after 76.76636242866516 seconds.
- figure4 is done after 79.38702249526978 seconds.
- figure5 is done after 35.066285610198975 seconds.
- figure6 is done after 72.53754782676697 seconds.
- figure7 is done after 42.55961585044861 seconds.

shak360:
Does the master branch use cached results? Or does it have access to more threads or something? Figure 1 is just a blank figure anway so the fact that it takes longer on my branch vs on master makes me think that this comparison is somewhat meaningless.

aarmey:
Check if the build ran on Aretha or Moriah. Moriah is almost twice as fast.

aarmey:
So my guess from this is that your branch makes Fig. 4 ~2X faster.

shak360:
Interesting, so `master` is building on Moriah but this branch is building on Aretha.

PR #274, opened 2020-03-20
Figure change

shak360:
Main figures (figures 3, 4, 5, 6) now use similar code. Similar code is stored in `figureCommon.py`. Changes are made to the rest of the code to allow for this.


shak360:
need to clean up `figureMaker` in `figureCommon.py` and then this should be done

PR #278, opened 2020-03-23
cleaning up figures

shak360:
final figure changes before working on fixing model bugs


shak360:
It'd probably be easier to just look at the figures on Jenkins to examine this PR. I removed the blank figure 1. I added figures 2 and 3 to be analyses on uncensored lineages (single lineages at a time with an increasing number of cells, and then multiple lineages with an increasing number of lineages). This brings the total figure count to 8.

shak360:
Will work on the captions soon.

PR #279, opened 2020-03-23
Adding more figures/captions

shak360:
Didn't necessarily need these new figures for the manuscript, but could possibly be used as a supplement. These figures would be used to formulate an argument on how to tackle the bias introduced by censorship. Some captions are changed. Old parts of the manuscript are removed.


shak360:
#279 is aalllmost done. Still need to add captions for figures 2,3,4,5 and some of the latex isn't rendering.

PR #281, opened 2020-03-24
Sv moretests

shak360:


shak360:
Added more figures; basically, we now have a censored and uncensored version of each figure. This is to help me figure out what's going wrong with the transition matrix estimation. Also State Distribution code, which we wanted to be the part that the user writes, is now no longer used in any of the model of lineage creation code (apart from one import statement). StateDistribution code is now in its own folder. Ultimately, these changes should make it easier to freeze or fix parameters and the model can solve for the other ones.

shak360:
Also, the MLE for the second gamma parameter is just the mean of the data divided by the first parameter, so that change is added (and is marginaly faster). I spent some time digging through scipy's implementation of gamma fit, hoping that their gamma fit function would be worth using. They have something called `_fitstart` which gets the gamma parameter by finding the skewness of the data (which is easy to do) but doesn't solve for the scale. They have a more general function that solves the MLE problem when both the shape and the scale are free, but this, unfortunately, uses `scipy.optimize.brentq`. However, there they use the the fact that the MLE for the scale parameter is  just the mean of the data divided by the shape parameter.

shak360:
Having double for loops in the censor code, where the second for loop did recursions from a parent to collect its entire tree always bothered me as a complex (yet textbook elegant) solution to what should be an easy problem. I rewrote the censoring to be just one for loop and now no longer involves any recursions. Hopefully, this works, but all it does is: first, if a cell's parent is censored, then the cell itself must be censored (regardless of observation). Second, if a cell isn't already censored, then check if it should be, and censor it. If it's a cell that should be time censored, then additionally cut its lifetime short and make its end time, the experiment time. I'm sure this works but would appreciate any tests that could be added. Ideally, we'd have a way to plot the resulting lineage and it'd be really easy to visually check if this was working. I'll work on adding tests for this in a later branch.

shak360:
Figures 1 and 2 are blank (I want these to be, as before, figures that describe the uncensored and censored synthetic lineages).

(estimation on uncensored lineages and populations) Figures 3 and 4 are working as expected (figure 3 transition matrix estimation could be better)

(estimation on censored lineages and populations) Figures 5 and 6 could be working a lot better. 

(estimation after varying state presence or proportion) Figure 7 (uncensored) is working as expected, sort of. Figure 8 (censored) could be working a lot better.

(estimation after varying similarity between states) Figure 9 (uncensored) is working as expected, sort of. Figure 10 could be working a lot better. 

I haven't really looked at the AIC figure.

aarmey:
What do you mean by "could be better"? That the estimation is wrong?

shak360:
image Figure 3

image Figure 4

shak360:
image Figure 5

image Figure 6

shak360:
@aarmey

PR #283, opened 2020-04-03
Added test

shak360:


lineage/tests/test_StateDistribution.py
```diff
@@ -78,7 +79,18 @@ def test_estimator(self):
         self.assertTrue(0.0 <= abs(estimator_obj.bern_p - self.E[0].bern_p) <= 0.1)
         self.assertTrue(0.0 <= abs(estimator_obj.gamma_a - self.E[0].gamma_a) <= 3.0)
         self.assertTrue(0.0 <= abs(estimator_obj.gamma_scale - self.E[0].gamma_scale) <= 3.0)
-
+        
+    def test_censor(self):
+        """
+        A unittest for testing whether censoring is working
```
aarmey, line 85: Can you say what the expectation is? It looks like if a parent is censored the daughters are?

PR #286, opened 2020-04-09
Jc hold constant

jclagarde:
Made changes to hold T and pi constant on figure 2

### DIFF
--- a/lineage/Analyze.py
+++ b/lineage/Analyze.py
@@ -98,7 +98,7 @@ def run_Analyze_over(list_of_populations, num_states, parallel=True, **kwargs):
         for _, prom in enumerate(prom_holder):
             output.append(prom.result())
     else:
-        for _, population in enumerate(list_of_populations):
+        for idx, population in enumerate(list_of_populations):
             output.append(Analyze(population, num_states, fpi=list_of_fpi[idx], fT=list_of_fT[idx], fE=list_of_fE[idx]))
 
     return output
--- a/lineage/Analyze.py
+++ b/lineage/Analyze.py
@@ -84,9 +84,9 @@ def run_Analyze_over(list_of_populations, num_states, parallel=True, **kwargs):
     list_of_populations: a list of populations that contain lineages
     num_states: an integer number of states to identify (a hyper-parameter of our model)
     """
-    list_of_fpi = kwargs.get('list_of_fpi', None)
-    list_of_fT = kwargs.get('list_of_fT', None)
-    list_of_fE = kwargs.get('list_of_fE', None)
+    list_of_fpi = kwargs.get('list_of_fpi', [None]*len(list_of_populations))
+    list_of_fT = kwargs.get('list_of_fT', [None]*len(list_of_populations))
+    list_of_fE = kwargs.get('list_of_fE', [None]*len(list_of_populations))
     output = []
     if parallel:
         exe = ProcessPoolExecutor()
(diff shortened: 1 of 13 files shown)


shak360:
image
This is what you get when you fix everything for figure 2, our simplest analysis. Uncensored, full populations of single lineages of increasing length.

shak360:
image
Fiing everything for figure 3, uncensored populations of increasing size (populations with more lineages).

shak360:
image

image

Figures 4 and 5 respectively. These figures are the same as above (2 -> 4) (3 -> 5) except that they are now censored by both time and fate.

lineage/BaumWelch.py
```diff
@@ -101,14 +101,16 @@ def fit(tHMMobj, tolerance=np.spacing(1), max_iter=200):
...
                 cell_groups[state].append(lineage[ii])
-
-        # population wide pi calculation
-        tHMMobj.estimate.pi = pi_estimate / sum(pi_estimate)
-        # population wide T calculation
-        tHMMobj.estimate.T = T_estimate / T_estimate.sum(axis=1)[:, np.newaxis]
-        # opulation wide E calculation
-        for state_j in range(num_states):
-            tHMMobj.estimate.E[state_j] = tHMMobj.estimate.E[state_j].estimator([cell.obs for cell in cell_groups[state_j]])
+        if tHMMobj.estimate.fpi is None:
```
aarmey, line 104: Why `None` and not true/false?

lineage/tHMM.py
```diff
@@ -32,9 +42,12 @@ def __init__(self, X, num_states):
             FOM (str): For now, it is either "E": Exponential, or "G": Gompertz
             and it determines the type of distribution for lifetime of the cells
         """
+        fpi = kwargs.get('fpi', None)
```
aarmey, line 45: You can just put `fT=None` in the arguments list to do the exact same thing.

lineage/tHMM.py
```diff
@@ -6,19 +6,29 @@
 
 
 class estimate:
-    def __init__(self, num_states):
+    def __init__(self, num_states, **kwargs):
+        self.fpi = kwargs.get('fpi', None)
+        self.fT = kwargs.get('fT', None)
```
aarmey, line 11: See my comment below.

lineage/Analyze.py
```diff
@@ -79,19 +84,22 @@ def run_Analyze_over(list_of_populations, num_states, parallel=True):
     list_of_populations: a list of populations that contain lineages
     num_states: an integer number of states to identify (a hyper-parameter of our model)
     """
+    list_of_fpi = kwargs.get('list_of_fpi', [None]*len(list_of_populations))
```
aarmey, line 87: This is the right way to handle this here, because the default is complex.

lineage/Analyze.py
```diff
@@ -21,10 +21,12 @@ def preAnalyze(X, num_states):
     --------
     tHMMobj {obj}:
     """
-
+    fpi = kwargs.get('fpi', None)
```
aarmey, line 24: Can just put this in the argument list with `fpi=None` and accomplish the same thing.

shak360:
So, in general, is that method preferred over using the `**kwargs` in the function defintiion argument line? @aarmey

aarmey:
@shak360 it is, if the defaults are going to be used within the function. If there are an unknown number of optional arguments, or most of the arguments are just being passed along to another function, then `**kwargs` is better.

PR #309, opened 2020-05-12
analyzing censorships independently

shak360:


shak360:
Some cleaning up I did with JC.

lineage/figures/figureCommon.py
```diff
@@ -19,17 +19,17 @@
...
-state0 = StateDistribution(0.99, 7, 7)
-state1 = StateDistribution(0.75, 7, 1)
+state0 = StateDistribution(0.98, 7, 7)
+state1 = StateDistribution(0.88, 7, 1)
 E = [state0, state1]
 
 # bern, exp_lambda
-state10 = expStateDistribution(0.99, 49)
-state11 = expStateDistribution(0.75, 7)
+state10 = expStateDistribution(1, 49)
```
aarmey, line 27: So no cells will die?

lineage/figures/figureCommon.py
```diff
@@ -19,17 +19,17 @@
 T = np.array([[0.9, 0.1], [0.1, 0.9]], dtype="float")
 
 # bern, gamma_a, gamma_scale
-state0 = StateDistribution(0.99, 7, 7)
-state1 = StateDistribution(0.75, 7, 1)
+state0 = StateDistribution(0.98, 7, 7)
+state1 = StateDistribution(0.88, 7, 1)
```
aarmey, line 23: Is it better to make these more similar?

lineage/states/StateDistribution.py
```diff
@@ -94,16 +94,16 @@ def gamma_estimator(gamma_obs):
     """
     N = len(gamma_obs)
 
-    xbar = (sum(gamma_obs) + 7e-10) / (len(gamma_obs) + 1e-10)
+    xbar = (sum(gamma_obs) + 49e-10) / (len(gamma_obs) + 1e-10)
```
aarmey, line 97: Maybe `np.finfo(np.float64).eps` would be useful here?

shak360: I wanted to change this in a different pull request, if that's OK with you.

PR #311, opened 2020-05-12
fix general issues with global estimation and gamma parameter estimation

shak360:


aarmey:
That is a BOLD claim for a PR title. :-)

shak360:
> 
> 
> That is a BOLD claim for a PR title. :-)

Yes haha, but I think this bug-fix hopefully warrants it. I think "FIXES (ALMOST) EVERYTHING!!" would be a more appropriate title.

shak360:
There's a TLDR at the bottom.

So we sorta knew that the code on `master` was "working" (tests were passing and figures were being made), but there was clearly something wrong that was hard to identify. Some things we noticed that were wrong were:

1. the transition matrix estimates were wrong (whether or not the other parameter estimates were correct)
2. behavior in estimating censored lineages didn't correlate with behavior with estimating uncensored lineages (besides just being worse)
3. performance of estimation HEAVILY relied on 2 things: initialization of the parameters themselves and default/error outputs that were hardcoded into our estimators.
4. during Baum-Welch, the cells would sometimes be grouped into all one state which should have been rarer than it was

Something I noticed while working on this were some more serious errors:

5. more than 50 iterations of Baum-Welch were being run which seemed like a lot for our given ideal cases
6. the likelihood that we were trying to maximize wasn't always being maximized (unless our estimators were configured to output desirable results with those hard-coded if-statements) even though EM guarantees it 

There were about two or three things that we thought could be the source of the error:

1. global estimation of transition matrices were wrong 
2. attribute poor estimation to censorship being a difficult to handle (but the EM algorithm _should_ be able to handle this to some degree because of proof-of-correctness?)
3. general bugs or inconsistencies with the model

---

Global estimation of the transition matrices happens in Baum-Welch during the M step. This was the source of the first error. For Baum-Welch, the E-step calculates parameters that are used to estimate T. Namely, zetas, which represent the expected frequency of transition between states, and gammas, which represent the expected state a cell is to be in at a given time point (time point in the Markov tree, not time point relative to experiment time). For each lineage, the ratio of these is taken for the elements of the transition rate matrix. This works for single lineages. However, for multiple lineages, these matrices were first evaluated and then averaged at the end. This doesn't work, because as Dr. Meyer noted, lineages with different frequencies (where a lineage has nearly all cells of one state) and different sizes can alter/bias the result. I thought the fix for this was weighting each of the matrices by some factor based on the lineage size or the number of transitions or the number of cells in that state, both of which were wrong. The reason for this is a really basic mathematical error: what we want is the ratio between the number of transitions across all lineages represented by the zetas divided by the number of times the starting point of that transition is that respective state represented by the gammas. If we let zetas_1, gammas_1, zetas_2, gammas_2, represent those expected values for two lineages, then what we were previously doing was:

```normalized[(zetas_1/gammas_1) + (zetas_2/gammas_2)]```

What we want is:

```normalized[(zetas_1 + zetas_2)/(gammas_1 + gammas_2)]```

We can see how the top clearly doesn't equal the bottom. This change is reflected in the code in Baum-Welch, where zetas and gammas are summed over the lineages and then divided, instead of summing the ratio between them and then normalizing.

---

The second major change was the second parameter of Baum Welch that is estimated in the M step,  the set of parameters that define the state distributions, that is, the Bernoulli p, the gamma a, and scale for each of the states. What we would originally do is, use our gamma variables (which as before, represent the expectation that a cell is in that state at a given time point), find which gamma for which state was larger using an argmax, group the cells into states based on which cells had the larger gamma, and then estimate the state parameters based on those selected cells. This creates two problems: gammas represent an expectation so turning these likelihoods into 1s or 0s might be a misinterpretation of what they mean and secondly, in every guide/lecture to Baum Welch, gammas (the expectation that a cell is in that state) is utilized in the estimation of the parameter itself, which we weren't doing. What gammas really represents is a sort of participation weighting or a factor that says how confident the model is in saying that that particular cell belongs to that state. I think component/state identifier or participation variable are better names for gamma. This variable is common in mixture models and in general, EM algorithms, but is never used in the way we implemented. What we were originally confused about, when we were writing Baum Welch, was how to convert how people used gammas in Baum Welch to our use case. Usually, everyone had these multinomial observations (like the particular weather on a day or someone's mood), so instead, we incorrectly wrote that argmax way of estimating the parameters.

It's hard to explain the rest using words, but lecture 11, slide 34 in this lecture http://www.utstat.toronto.edu/~rsalakhu/sta4273/notes/Lecture11.pdf explains it in math. 

As an example, the Bernoulli MLE estimator is just the average of the data, the sum of the observations over the length. But because the gammas for each cell for each state represents the confidence that that cell is in that state, we can instead let the Bernoulli estimator see ALL the data but weight it by the gammas. The modified estimator then becomes the sum of gammas times the observations over the sum of the gammas. If we consider the case where the gammas are all one, for one of the states, then we see that the estimator reduces to the normal estimator without the gammas (because you're just multiplying every observation by 1 and the sum of the 1s is just the number of observations). If gammas is 1 for half of the cells for one state, and 1 for half of the cells for the other state, we see that its the same as the estimator seeing ONLY those cells for those states (because the gammas being 0 would drop some observations out). This change is reflected in Baum Welch and in the individual estimators.

---

The Bernoulli and exponential likelihood are easy to adapt with the gammas because their MLEs are just the sample averages. The gamma estimator is a little more difficult. I did the math on paper and wrote a Halley root finder using the first, and second-order derivatives of the likelihood incorporated with the gamma weights. I couldn't find an answer for this online, but I'm pretty sure its correct and it works. I think it works because

1. we don't need default outputs to move our parameter estimates around when they're bad or nonsensical
2. it isn't heavily reliant on the initial value of the parameters
3. can now see all observations instead of just some of them and no longer runs into the case of an empty list
4. _any_ pair of gamma values are well estimated instead of hand-selected ones
5. doesn't utilize an iterative Newton's method

Some issues I did have were
1. when we analyze SINGLE lineages of CENSORED cells with fewer than 50-100 cells, the parameter estimates are wildly wrong (in the 1000s) for the first iterationof Baum Welch, so there is that one case where I have to hard-code an if-statement to return a more sensible parameter estimate. You can see this in the new gamma estimator. The problem of having the estimator return a parameter estimate for gamma a and scale like (1000, 0.09), is that the pdf will then return 0.0 for most observations. When this happens, I ask that the model be rerun, but I don't know if the way I coded this was correct. 

I don't see this issue on any other type of lineage data, like populations of censored lineages (see below). I had some issues with figure 10, but I think that figure can be fixed on a different branch.

---
**TLDR**:
1. transition matrices are globally estimated
2. the gamma variable that our model estimates (which I'll call the component identifier variable or the participation variable) which is the expectation that a cell is in that state, is now being correctly utilized in the estimation of the state distribution parameters
3. bernoulli, gamma, and exponential estimators now utilize the participation variable in their estimation, and are rewritten
4. gamma estimator is written as a halley root finder utilizing the participation variable


Some TODOs:
I think it'll be easier to handle censorship because I have more confidence that the model as a whole is working (that is in the way Mr. Baum or Mr. Welch would have wanted it). Knowing that now, we can look at the figures of the censored lineages with a different perspective.

image

The above is the new figure 15, POPULATIONS of CENSORED lineages with exponential times. Something to note is how the model estimates the exponential parameters which is now biased almost perfectly by a fixed amount lower than the true value. Something important to note is that the exponential estimator NO LONGER has a default output, nothing to output from a hard-coded if statement. This makes me believe that censorship can be effectively analyzed. I have a plan, but will approach this in a new branch.

For comparison, here's the new figure 5, POPULATIONS of CENSORED lineages with gamma times.
image

It seems like here, the estimator is more likely to bias the shape down, instead of moving the scale up. I have looked into gamma estimators for shapes that are equal across components or states, but many utilize a Newton solver. What we could do, is just force the model to share the parameter for both states, but I don't know if that's the MLE or optimal way.

Here are the uncensored versions of the figures above:

image

image

File: lineage/states/StateDistribution.py [hunk 1]
```diff
@@ -36,12 +37,19 @@ def pdf(self, tuple_of_obs):  # user has to define how to calculate the likeliho
...
-        gamma_ll = gamma_pdf(tuple_of_obs[1], self.gamma_a, self.gamma_scale)
+        try:
+            bern_ll = bern_pdf(tuple_of_obs[0], self.bern_p)
+        except ZeroDivisionError:
+            assert False, f"{tuple_of_obs[0]}, {self.bern_p}"
+            
+        try:
+            gamma_ll = gamma_pdf(tuple_of_obs[1], self.gamma_a, self.gamma_scale)
+        except ZeroDivisionError:
+            assert False, f"{tuple_of_obs[1]}, {self.gamma_a}, {self.gamma_scale}"
```
shak360, line 48: @aarmey I want this to be caught in `Analyze` where we have it set up to run our model multiple times in case it catches an AssertionError. However, the error we catch here is a ZeroDivisionError which I "convert" to a AssertionError so that the model reruns. This is because on **really rare** cases (like the first iteration of the model is just really bad and leads to bad parameter estimates which leads to a pdf evaluating to 0), instead of rerunning, the code throws a ZeroDivisionError which doesn't cause the model to rerun even though we want it to. Is this the way to do it? I'm referring to this line: https://github.com/meyer-lab/lineage-growth/blob/869d16bde9ad866e17669c014c4cccf414ad5a38/lineage/Analyze.py#L29

lineage/states/StateDistribution.py
```diff
@@ -87,25 +97,28 @@ def __repr__(self):
...
 
-    if b_hat < 1.0 or 50.0 < a_hat < 5.0:
-        return 7, 7
+    scale_hat = (sum(gammas*gamma_obs))/a_hat/sum(gammas)
+    
+    if a_hat > 100:
+        return 7, 3+(1*(np.random.uniform()))
 
-    return a_hat, b_hat
+    return a_hat, scale_hat
```
shak360, line 121: example of how the `gammas` or the participation variable, or the component or state identifier variable is now utilized in the estimation of the parameters themselves.

halley root finder is now used.

lineage/states/stateCommon.py
```diff
@@ -12,16 +12,15 @@ def bern_pdf(x, p):
...
 
 
-def bernoulli_estimator(bern_obs):
+def bernoulli_estimator(bern_obs, old_params, gammas):
     """
     Add up all the 1s and divide by the total length (finding the average).
     """
-    bern_obs = np.append(bern_obs,1)
-    return (sum(bern_obs)) / (len(bern_obs))
+    return sum(gammas*bern_obs) / sum(gammas)
```
shak360, line 23: example of how the participation variable is used in the Bernoulli parameter estimation

lineage/BaumWelch.py
```diff
@@ -73,44 +73,42 @@ def fit(tHMMobj, tolerance=np.spacing(1), max_iter=200):
...
+            # local pi estimate
             pi_estimate += gamma_array[0, :]
 
-            denom = get_all_gammas(lineageObj, gamma_array)
-            numer = get_all_zetas(
+            # local T estimate
+            numer_estimate += get_all_zetas(
                 lineageObj=lineageObj, beta_array=betas[num], MSD_array=tHMMobj.MSD[num], gamma_array=gamma_array, T=tHMMobj.estimate.T
             )
+            denom_estimate += get_all_gammas(lineageObj, gamma_array)
```
shak360, line 95: The values are collected and summed over the lineages first, before their ratio is taken

lineage/states/stateCommon.py
```diff
@@ -12,16 +12,15 @@ def bern_pdf(x, p):
     probability distribution function.
     """
     # bern_ll = self.bern_p**(tuple_of_obs[0]) * (1.0-self.bern_p)**(1-tuple_of_obs[0])
-
+    
     return (p**x) * ((1.0 - p)**(1 - x))
 
 
-def bernoulli_estimator(bern_obs):
+def bernoulli_estimator(bern_obs, old_params, gammas):
```
aarmey, line 19: Is `old_params` used?

lineage/tests/test_Viterbi.py
```diff
@@ -27,7 +27,7 @@ def test_vt(self):
         E = [state_obj0, state_obj1]
 
         X = LineageTree(pi, T, E, desired_num_cells=(2 ** 11) - 1)
-        tHMMobj, pred_states_by_lineage, _ = Analyze([X], num_states=2)
+        tHMMobj, pred_states_by_lineage, LL = Analyze([X], num_states=2)
```
aarmey, line 30: Isn't `LL` unused?

lineage/states/stateCommon.py
shak360: It's not in this estimator, and not in the exponential one either, but it is in the gamma estimator. I wanted it there so that all the estimators follow the same argument line:
1. observations
1.5 censor observations (for the time distributions)
2. a tuple of old params,
3. gammas

I can remove it if it's better that it's not there.

lineage/tests/test_Viterbi.py
shak360: Fixed, was an artifact of debugging.

lineage/states/StateDistribution.py
aarmey: Couldn't you catch the `ZeroDivisionError` in Analyze?

shak360: Would it just be:

`except AssertionError or ZeroDivisionError:`

shak360: Never mind, https://stackoverflow.com/questions/6470428/catch-multiple-exceptions-in-one-line-except-block I see how to do it.

lineage/states/StateDistribution.py
aarmey: Aren't you only doing one iteration of Halley here? It's iterative and so would need to run until convergence.

If you pass the `fprime2` argument here, scipy can take care of Halley's method and convergence checks:

https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.newton.html#scipy.optimize.newton

lineage/states/stateCommon.py
aarmey: Ok. Good as is.

lineage/states/StateDistribution.py
shak360: Fixed now.

lineage/states/StateDistribution.py
shak360: I tried using the Newton root finder originally (it was the documentation for that that led me to Halley's method) but the issue I had with it was it would create a Runtime error when it wasn't converging. When I tried it in a notebook I was getting the Halley's method to converge in 1 to 2 iterations, so I thought this would be better.

lineage/states/StateDistribution.py
```diff
@@ -87,25 +97,27 @@ def __repr__(self):
...
+    def fprime2(k, s): return (2*((1/k)-sc.polygamma(1, k))**2)/((np.log(k)-sc.polygamma(0, k))**3) + (sc.polygamma(2, k)+(1./k**2))/((np.log(k)-sc.polygamma(0, k))**2)
 
-    xbar = (sum(gamma_obs) + 49e-10) / (len(gamma_obs) + 1e-10)
-    x_lnx = [x * np.log(x) for x in gamma_obs]
-    lnx = [np.log(x) for x in gamma_obs]
-    # gamma_a
-    a_hat = (N * (sum(gamma_obs)) + 7e-10) / (N * sum(x_lnx) - (sum(lnx)) * (sum(gamma_obs)) + 1e-10)
-    # gamma_scale
-    b_hat = xbar / a_hat
+    def halley(k, s): return k - ((2 * (f(k, s) * fprime(k, s))) / (2 * (fprime(k, s)**2) - f(k, s) * fprime2(k, s)))
```
aarmey, line 112: Isn't this only doing one iteration of Halley?

shak360: Yep but I was getting it to the answer within one to two iterations so I wasn't sure if a loop was necessary.

shak360: The second iteration was always a very small number away from the first iteration (around ~0.0005)

aarmey: Really think you still need to iterate... there could easily be pathological cases.

aarmey: If you're having trouble with this I'm fine reverting to a version that's working (one iteration) and then just opening an issue to revisit this later.

shak360: Yeah I can't seem to figure it out, and I like approaching the problem on a new issue/branch later.

lineage/figures/figureCommon.py
```diff
@@ -20,12 +20,12 @@
...
 # bern, gamma_a, gamma_scale
 state0 = StateDistribution(0.99, 7, 7)
-state1 = StateDistribution(0.75, 7, 1)
+state1 = StateDistribution(0.88, 7, 1)
 E = [state0, state1]
 
 # bern, exp_lambda
-state10 = expStateDistribution(0.99, 49)
-state11 = expStateDistribution(0.75, 7)
+state10 = expStateDistribution(1, 49)
```
aarmey, line 27: Do you mean for both states to be 1?

shak360: I just changed them to the original parameters.

PR #325, opened 2020-05-18
Linting and gamma estimator

aarmey:
This should fix the gamma estimator. A region of the function was leading to NaN derivatives, so I switched to `brentq` for root finding.

### DIFF
--- a/lineage/states/StateDistribution.py
+++ b/lineage/states/StateDistribution.py
@@ -4,6 +4,7 @@ import numpy as np
 import scipy.stats as sp
 from numba import njit
 import scipy.special as sc
+from scipy.optimize import brentq
 
 
 from .stateCommon import bern_pdf, bernoulli_estimator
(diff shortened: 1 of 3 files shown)


lineage/states/StateDistribution.py
```diff
@@ -102,22 +103,16 @@ def gamma_estimator(gamma_obs, gamma_censor_obs, old_params, gammas):
...
-    
-    
-    a_hat = a_hat_list[-1]
-    scale_hat = (sum(gammas * gamma_obs)) / a_hat / sum(gammas)
-
-    if a_hat > 100:
-        return 7, 3 + (1 * (np.random.uniform()))
+    gammaCor = sum(gammas * gamma_obs) / sum(gammas)
+    s = np.log(gammaCor) - sum(gammas * np.log(gamma_obs)) / sum(gammas)
+    def f(k): return np.log(k) - sc.polygamma(0, k) - s
```
shak360, line 108: Is there an advantage to using `lambda` vs `def` for defining functions?

aarmey: Not that I know of.

PR #328, opened 2020-05-19
Use copies of the estimates when shuffling their elements instead of doing it in-place

shak360:
Python error where I thought creating a temporary copy of the T matrix would create a different T to alter, but it was actually changing the original T. 

Now uses `np.copy()`.

### DIFF
--- a/lineage/Analyze.py
+++ b/lineage/Analyze.py
@@ -168,11 +168,7 @@ def Results(tHMMobj, pred_states_by_lineage, LL, fpi, fT, fE):
         for col_idx in range(tHMMobj.num_states):
             temp_T[row_idx, col_idx] = tHMMobj.estimate.T[switcher_map[row_idx], switcher_map[col_idx]]
 
-    results_dict["switched_transition_matrix"] = temp_T
-    if fT is None:
-        results_dict["transition_matrix_norm"] = np.linalg.norm(temp_T - tHMMobj.X[0].T)
-    else:
-        results_dict["transition_matrix_norm"] = np.linalg.norm(tHMMobj.estimate.T - tHMMobj.X[0].T)
+    results_dict["transition_matrix_norm"] = np.linalg.norm(temp_T - tHMMobj.X[0].T)
         
     # Rearrange the values in the pi vector
     temp_pi = np.zeros((tHMMobj.num_states))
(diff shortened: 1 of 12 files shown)


lineage/Analyze.py
```diff
@@ -164,33 +163,42 @@ def Results(tHMMobj, pred_states_by_lineage, LL):
...
-    results_dict["transition_matrix_norm"] = np.linalg.norm(temp_T - tHMMobj.X[0].T)
-
+    if fT is None:
+        results_dict["transition_matrix_norm"] = np.linalg.norm(temp_T - tHMMobj.X[0].T)
+    else:
+        results_dict["transition_matrix_norm"] = np.linalg.norm(tHMMobj.estimate.T - tHMMobj.X[0].T)
+        
     # Rearrange the values in the pi vector
-    temp_pi = tHMMobj.estimate.pi
+    temp_pi = np.zeros((tHMMobj.num_states))
```
shak360, line 178: This line is one of the issues where copying the estimate vector and then changing the copy also changed the original (which isn't what we want).

lineage/Analyze.py
```diff
@@ -164,33 +163,42 @@ def Results(tHMMobj, pred_states_by_lineage, LL):
     results_dict["switcher_map"] = switcher_map
 
     # Rearrange the values in the transition matrix
-    temp_T = tHMMobj.estimate.T
+    temp_T = np.zeros((tHMMobj.num_states,tHMMobj.num_states))
```
shak360, line 166: The other time where I copy the estimate T and then change the copy, but this also changes the original. Now it uses a matrix that is no where related to the original copy, so changing it doesn't affect the original.

shak360:
So on `master`, if we try to make a figure when fixing T and pi, we get the following:

image

Note how the errors for the T matrix norm and the pi vector norm _should_ be zero (because they're fixed) but aren't. This is because of the errors above. After this fix, on this branch this figure looks like the following:

image

lineage/Analyze.py
```diff
@@ -96,7 +96,7 @@ def run_Analyze_over(list_of_populations, num_states, parallel=True, **kwargs):
     return output
 
 
-def Results(tHMMobj, pred_states_by_lineage, LL):
+def Results(tHMMobj, pred_states_by_lineage, LL, fpi, fT, fE):
```
shak360, line 99: The Results function should know if we have fixed parameters that were given in analysis.

lineage/Analyze.py
```diff
@@ -221,15 +229,15 @@ def Results(tHMMobj, pred_states_by_lineage, LL):
     return results_dict
 
 
-def run_Results_over(output):
+def run_Results_over(output, list_of_fpi, list_of_fT, list_of_fE):
```
shak360, line 232: The 'Results' function should know if we have fixed parameters.

lineage/Analyze.py
shak360: Outdated, ignore.

lineage/Analyze.py
shak360: Outdated, ignore.

lineage/Analyze.py
aarmey: You want `np.copy()`

https://numpy.org/doc/1.18/reference/generated/numpy.copy.html

PR #329, opened 2020-05-21
Scott tutorial

scottdtaylor95:
I believe this should be complete. The "guide" is fairly short/simple, as I feel like most of it is self-explanatory or already covered in the Overview.


shak360:
Looks good to me, I can render the math in the comments using LaTeX on a different branch.

PR #330, opened 2020-05-22
phase durations as observations

Farnazmdi:
(1) Brought back the stateDistribution file we had for phases as separate observations and modified according to the new code. It is in the states folder and the name is `stateDistPhase.py`.
(2) Made a notebook with the same structure as the `0_overview` with phase durations separately as observations. 
(3) Made figure16 as a duplicate of figure 6 with the new state distribution.

Overall looks working, but the accuracy of parameter estimation could be improved.
Here is the figure 16:
image


lineage/figures/figure16.py
```diff
@@ -0,0 +1,185 @@
...
+
+        population.append(tmp_lineage)
+
+        # Adding populations into a holder for analysing
+        list_of_populations.append(population)
+        list_of_fpi.append(pi)
+        list_of_fT.append(T)
+        list_of_fE.append(E)
+
+    return commonAnalyze(list_of_populations, xtype="prop", list_of_fpi=list_of_fpi)
```
shak360, line 185: Are you sure you want `xtype="prop"` here @Farnazmdi? This measures the percentage of cells in state 0 vs the total. The default type measures the length or the number of cells in the population, and there's also `wass` which collectes the Wasserstein distance.

Farnazmdi: Changed it to `length`, thanks. So length is the default?

shak360: Yea so that function:

https://github.com/meyer-lab/lineage-growth/blob/5f97dff2c04282a46737456e1f2a957539db6aef/lineage/figures/figureCommon.py#L78

You can actually leave out `xtype="length"` in the function call and it should work I think

Farnazmdi: Is this what you expect to see?
image

aarmey:
Review with Shakthi then tag me for review.

lineage/figures/figure16.py
```diff
@@ -0,0 +1,185 @@
...
+
+        population.append(tmp_lineage)
+
+        # Adding populations into a holder for analysing
+        list_of_populations.append(population)
+        list_of_fpi.append(pi)
+        list_of_fT.append(T)
+        list_of_fE.append(E)
+
+    return commonAnalyze(list_of_populations, xtype="length", list_of_fpi=list_of_fpi)
```
shak360: Yea so for this figure (I think this is figure 6 right), each lineage has the same number of cells (which is why all the data points are on the same number). The only thing that differs from each lineage is the transition matrix (line 164). I would recommend copying like Figure 5 (GROUPS of CENSORED lineages) or Figure 3 (GROUPS of UNCENSORED lineages) and then using the state distribution.

shak360:
@Farnazmdi I'd recommend doing this: https://github.com/meyer-lab/lineage-growth/blob/5f97dff2c04282a46737456e1f2a957539db6aef/lineage/figures/figureCommon.py#L13

You can import your state distribution like so, create an emission object, and then import it in a figure file. You can make a duplicate of figure 3 or 5, using that emissions object, the same way figures 13 and 15 are duplicates of figure 3 and figure 5 using the exponential state distribution object.

Farnazmdi:
In the notebook with the same parameter it works and estimates relatively good, but in the figures it throws assertion errors suggesting to change the estimator! What do you mean by changing the estimator?

shak360:
> 
> 
> In the notebook with the same parameter it works and estimates relatively good, but in the figures it throws assertion errors suggesting to change the estimator! What do you mean by changing the estimator?

It's hitting a runtime error; try making it so the lineages aren't censored and it should be fine, I think.

Farnazmdi:
Could you tell me how to do that?

Farnazmdi:
Got it, it's censor_condition in lineage_tree, right?

shak360:
Yea but they are uncensored by default, that is, you don't need to explicitly `state censor_condition=0`.

I would still recommend adding your emissions object the way its done with the exponential state distribution in `StateCommon.py`. and then copying figure 3 or figure 13 and then adding your emissions object to your copy of the figure.

Farnazmdi:
I did what you recommended in figure 5.
Here is the simulation:
image

Farnazmdi:
@shak360 Could you review to see if you have other comments?

shak360:
Yea if you don't mind naming it figure 25 (because it's like the analogue of figure 15 and figure 5).

Farnazmdi:
image

lineage/states/stateCommon.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, startT, endT):
         self.endT = endT
 
 
-def assign_times(lineageObj):
+def assign_times(lineageObj, *kwargs):
```
shak360, line 39: I think ultimately, this `assign_times` function should be `StateDistribution` specific. @scottdtaylor95 had some trouble with it when he only had Gaussian observations. We can do this on a later branch though.

shak360: See #335 (can be fixed on a later branch).

lineage/figures/figureCommon.py
```diff
@@ -247,3 +252,96 @@ def figureMaker1(ax, x, paramEst, accuracies, tr, pii, paramTrues, xlabel="Numbe
     ax[i].set_title("Initial Probability Matrix Estimation")
     ax[i].grid(linestyle="--")
     ax[i].tick_params(axis="both", which="major", grid_alpha=0.25)
+
+def figureMaker2(ax, x, paramEst, accuracies, tr, pii, paramTrues, xlabel="Number of Cells"):
```
aarmey, line 256: If this is specific to one figure, it should just be in that file.

Farnazmdi: This is for the figures we want to make when we have G1 and G2 separated.
Should I move only this function back to the figure file or the parameters, too?

PR #340, opened 2020-05-27
Add function for placing cartoons

aarmey:
Here is the function for placing cartoons within a figure. Note that this has to happen **after** saving the SVG file, and so the best place to use this is in the `genFigures.py` script as demonstrated [here](https://github.com/meyer-lab/gc-cytokines/blob/master/genFigures.py#L24).


jclagarde:
I added the cartoons svg to the Git repo within the 'data' folder. It sent as a pull request. Do I need to wait for it to be approved before using genFigures.py?

aarmey:
I'll review it right now, and merge this, so that we have everything on master.

PR #342, opened 2020-05-28
Create cartoons

jclagarde:
Added cartoons svg in data folder


jclagarde:
ah, I had put all of the figure svg files into one. I will add them separately right now

PR #356, opened 2020-05-28
An attempt at gamma censoring

aarmey:
This is the approach I thought would work—to weight by gamma, and then fit the likelihood using the survival function for censored observations. It clearly doesn't. I'd be interested in your thoughts.

### DIFF
--- a/lineage/states/StateDistribution.py
+++ b/lineage/states/StateDistribution.py
@@ -4,7 +4,7 @@ import numpy as np
 import scipy.stats as sp
 from numba import njit
 import scipy.special as sc
-from scipy.optimize import brentq
+from scipy.optimize import brentq, minimize
 
 
 from .stateCommon import bern_pdf, bernoulli_estimator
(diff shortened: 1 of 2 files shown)


aarmey:
I’m at least still missing the effect of censoring in the expectation step. Will try that shortly.

lineage/states/StateDistribution.py
```diff
@@ -38,15 +36,16 @@ def pdf(self, tuple_of_obs):  # user has to define how to calculate the likeliho
...
-        except ZeroDivisionError:
-            assert False, f"{tuple_of_obs[0]}, {self.bern_p}"
+        bern_ll = bern_pdf(tuple_of_obs[0], self.bern_p) if tuple_of_obs[2] == 1 else 1.0
 
         try:
-            gamma_ll = gamma_pdf(tuple_of_obs[1], self.gamma_a, self.gamma_scale)
+            if tuple_of_obs[2] == 1:
+                gamma_ll = gamma_pdf(tuple_of_obs[1], self.gamma_a, self.gamma_scale)
+            else:
+                gamma_ll = sp.gamma.sf(tuple_of_obs[1], a=self.gamma_a, scale=self.gamma_scale)
```
shak360, line 45: I agree with/like this approach. Knowing which observations are censored should be used when calculating the likelihood of that observation.

PR #362, opened 2020-05-29
small fix with assigning states to new cells using multinomial rv

shak360:
image

### DIFF
--- a/lineage/CellVar.py
+++ b/lineage/CellVar.py
@@ -165,7 +165,7 @@ def double(parent_state, T):
 
     # The index of the 1 value in the resulting list in the two trials represent the state that was picked for that
     # dice toss.
-    return left_state_results.index(1), right_state_results.index(1)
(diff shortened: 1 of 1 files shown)


lineage/CellVar.py
```diff
@@ -165,7 +165,7 @@ def double(parent_state, T):
 
     # The index of the 1 value in the resulting list in the two trials represent the state that was picked for that
     # dice toss.
-    return left_state_results[1], right_state_results[1]
```
shak360, line 168: Though this is faster than the proposed method, this method only matches the results we want when our lineage tree only has two states. When we have more than three states, this won't work, and a lot of cells will be made 0 inccorrectly.

shak360: image

For example, the existing method would not be able to say that the cells should be state 2 for some of these possible transition matrix rows with 3 states.

PR #363, opened 2020-05-29
Update pytest.yml

shak360:


shak360:
I'm trying to make `pytest --cov=lineage` ignore the figures folder but I'm not sure if this is the way to do it @aarmey .

aarmey:
This looks right as far as I can tell...

.github/workflows/pytest.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
     - name: Test with pytest
       run: |
-        pytest --cov=lineage --cov-report=xml
+        pytest --cov=lineage --cov-report=xml --cov-config=.coveragerc
```
aarmey, line 33: With the way you have it, this should be `--cov-config=.github/workflows/coveragerc`

shak360:
Cool, I think it's working as intended now.

PR #365, opened 2020-05-30
Clean state interface

aarmey:
This has the state distributions update in-place, and stores the parameters only in the `params` vector, rather than in two places. Finally, this moves the zip/list/map/list step outside of the StateDistribution, so that the interface is cleaner.

### DIFF
--- a/lineage/BaumWelch.py
+++ b/lineage/BaumWelch.py
@@ -71,7 +71,7 @@ def calculateQuantities(tHMMobj):
     return NF, betas, gammas, LL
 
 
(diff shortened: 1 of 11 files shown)


aarmey:
There's a regression in the fitting performance, I think because I changed the values the distributions are initialized to. Should be a quick fix.

aarmey:
Well, the first step in BW should be random state assignment and then estimation. If this is the case the StateDistribution initialization shouldn't matter.

I found a way to avoid copying. Will merge this.

Issue #339, opened 2020-05-27
The state distributions should be directly tested


aarmey:
This can simply be generating a series of observations and checking you get an accurate parameter estimation out. You could also check that if `gamma` is reduced for all the observations identically, that you get the same answer.

aarmey:
A good way to do this for all StateDistribution types would be create a bunch of cells using the default initializer, then fit these and check you get an estimate close to the real center.

Farnazmdi:
Handling empty observations means that if the observation arrays are empty, it should return some error?

aarmey:
Actually, you can ignore that about empty observations. Shakthi and I ended up discussing that in a PR, and decided we can set it up so that estimators can assume they will always get observations.

Farnazmdi:
I see. OK, so I will add the tests from the other issue and tag Shakthi and then you for review.

PR #377, opened 2020-06-01
tests added

Farnazmdi:
Addressed issue #339. The tests now include all the state distributions and they pass.


aarmey:
Is this organized as well as it could be? I'm not sure I follow how it's organized into tests that address one potential issue each. If @shak360 disagrees with me, though, then I'm ready to merge.

Farnazmdi:
For the 3 state model check, I only added extensions to the existing tests in tHMM.py and Bawm-Welch.py
It follows the same pattern as the 2-state model follows.
For the issue to address state distributions, I added the state distribution from Gaussian and G1/G2 again into the existing test in test_StateDistribution. @aarmey

aarmey:
Ok, I feel like there might be a better organization strategy, but that's separate from this issue.

PR #381, opened 2020-06-03
minor changes

shak360:


shak360:
@aarmey I re-implemented a censored exponential estimator combined with your idea of having the censored likelihood (survival function) be used in the pdf function. It works better than just using the censored exponential estimator without the survival function being used. I tried doing this with the gamma state distribution, but I can't seem to get the minimizer to behave properly. For exponential, the analytical estimator is easily derived. For gamma, the survival function (1 minus the CDF) uses something called the lower incomplete gamma function, which is tough to deal with analytically, especially because it's being subtracted by 1 inside a logarithm. It's not that bad though, because multiplying the log sf and the log pdf by their respective gammas is what we're trying to calculate anyways. Do you have any ideas on how to proceed?

aarmey:
Looks great. Why don't you revert the gamma estimator changes, then we can merge this? It looks like what you have for the gamma estimator is roughly what I had there before, just in different form.

I think I saw roughly the same thing—the gamma estimator behaving poorly. We should be able to get the minimization approach working, but I haven't looked at it carefully. Probably the best approach would be to set `minimize()` to verbose output, and watch how it's behaving. If we're getting `NaN` or `Inf`, we can look at where those arise from.

This is also why I was thinking to look at censorship with a one state model. I think sometimes the `gamma` is the source of the problems. Also, we then would know that estimator convergence is the only issue.

Issue #368, opened 2020-05-30
Go through the documentation warnings on build


aarmey:
I think some of the documentation is formatted incorrectly for Sphinx.

aarmey:
Also, math in the docstrings should be formatted as:

```
:math:`equation`
```

PR #383, opened 2020-06-03
modified docstrings

Farnazmdi:
#368


Farnazmdi:
For the existing docstring, I edited them based on [this link](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html). It is more time consuming than I thought to do all of them. If it is alright, JC or Scott if they want to become more familiar with the code, they could follow the pattern of these files I have edited and do the rest of the files. If not, I can do them.
Please let me know what you think @aarmey. Thanks.

aarmey:
P.S. Make sure this doesn't close the issue, then.

Issue #370, opened 2020-05-30
Make one-state evaluation of censorship


aarmey:
This should make evaluating what's wrong much easier, since we don't need to worry about state assignment.

Issue #357, opened 2020-05-29
Convert formatted messages to f-strings


aarmey:
This will make the error reporting code easier to read.

aarmey:
@scottdtaylor95, @shak360 can explain this.

PR #387, opened 2020-06-05
allow creation/fitting/testing of one state lineages

shak360:


shak360:
Fixes #370

shak360:
fixes #357

shak360:
@aarmey Let me know if this is what you were expecting. The test file isn't being tested when pytest is run in the checks, but I was thinking it's just waiting to be pushed first or something.

lineage/CellVar.py
```diff
@@ -158,7 +158,7 @@ def double(parent_state, T):
 
     # Rolling two of the same loaded dice separate times and assigning
     # where they landed to states
-    left_state_results, right_state_results = sp.multinomial.rvs(n=1, p=np.squeeze(T[parent_state, :]), size=2)
```
shak360, line 161: Only thing that needed to be changed to allow for one state lineages.

lineage/tests/test_StateEstimation.py
```diff
@@ -0,0 +1,81 @@
+""" Unit test file. """
```
shak360, line 1: New unit test file with one state lineages/fitting and evaluation of parameter estimation.

aarmey:
Looking into why the tests don't run...

aarmey:
@shak360 OH! Methods that should run within the test class must start with `test_` in their name. Look at the other files to see what I mean.

shak360:
Ah, got it. I'll make the change in a new branch.

PR #391, opened 2020-06-07
prevent test failures by checking lineages

shak360:


shak360:
Tests should fail a lot less now. The lineages just needed to be checked before they were fit (sometimes lineages would have 5 or fewer cells or 0 cells) and should have been remade before being tested.

PR #392, opened 2020-06-08
Refactoring

shak360:
Changes:
- mostly linting (adding docstrings, removing whitespace, changing import order, line lengths)
- made EL (emission likelihoods) a property of tHMM with getter and setter methods (simplifies calling of functions)
- made tHMM more like sklearn (has a `.predict()` method and  `.log_score()` method)

So a user can now do

```python
X = [LineageTree(pi, T, E, 2**9)]
solver = tHMM(X, 2)
solver.fit()
pred_states = solver.predict()

# log likelihood of predicted states using the fit estimated parameters
pred_log_prob_estimated_parameters = solver.log_score(pred_states) 

# log likelihood of predicted states using true parameters
# this is always less than the above
# pred_log_prob_true_parameters < pred_log_prob_estimated_parameters 
pred_log_prob_true_parameters = solver.log_score(pred_states, pi=pi, T=T, E=E)

# log likelihood of true states using the fit estimated parameters
# true_log_prob_estimated_parameters < pred_log_prob_estimated_parameters 
true_log_prob_estimated_parameters = solver.log_score([[cell.state for cell in X[0].output_lineage]]
```

- the `.predict()` method returns a state assignment list (just moved all the Viterbi commands to one function)
- the  `.log_score()` method returns a log-likelihood given a possible state assignment 
- the Viterbi test used to use something like the `.log_score()` function previously called `LLfunc()` but it used to fail sometimes. Cleaning up that function (which is now  `.log_score()`) makes the test not fail anymore. I think previously it was comparing lists instead of scalar values, but I'm not sure why it was failing often before.

Something I wanted to bring your attention to Dr. Meyer @aarmey , was that the log probability of the predicted states via Viterbi using the fit estimated parameters via Baum-Welch would have a **higher likelihood** than the true states using the true parameters, **always**. The difference isn't high, but it would still always be higher. I'm trying to wrap my head around if this is to be expected, but I'm not sure.


lineage/states/stateCommon.py
```diff
@@ -127,14 +127,22 @@ def get_experiment_time(lineageObj):
 
 
 def basic_censor(cell):
+    """
+    Censors nuclear family of a cell.
```
aarmey, line 131: Do you mean descendants?

lineage/tHMM.py
```diff
@@ -52,9 +56,24 @@ def __init__(self, X, num_states, fpi=None, fT=None, fE=None):
...
         self.fE = fE
-        self.X = X  # list containing lineages, should be in correct format (contain no NaNs)
-        self.num_states = num_states  # number of discrete hidden states
+        self.X = X  # list containing lineages
+        self.num_states = num_states  # number of discrete hidden states, should be integral
         self.estimate = estimate(self.X, self.num_states, fpi=self.fpi, fT=self.fT, fE=self.fE)
+        self.EL = get_Emission_Likelihoods(self)
+
+    # getter method
+    def get_EL(self):
```
aarmey, line 65: What's the benefit of this over direct access?

aarmey:
A *little* higher likelihood than the true answer makes sense, because the fit is to the actual samples, not the truth. Essentially this is the source of variance in the bias-variance tradeoff.

lineage/tHMM.py
shak360: Nothing, I thought it'd be the right way to do it, but you're right, I'd rather prefer the cleaner/shorter way to do it.

PR #399, opened 2020-06-13
figure 24 for phenotypes

Farnazmdi:
image


aarmey:
Don't the histogram and violin plots for the gamma distribution essentially show the same thing? Maybe the violin plot is more consistent?

Farnazmdi:
image

Farnazmdi:
Do the two plots for Bernoulli look good? Or I can merge them into one plot so that the pairs of the total, state 1, and state 2 are beside each other, and all are in one plot.

PR #403, opened 2020-06-20
Plot3

Farnazmdi:
Plot for phenotypes. It still needs some modifying for identifying the shape markers, but I guess it should be ready to merge for now.


Farnazmdi:
image

aarmey:
Have @shak360 review.

lineage/Analyze.py
```diff
@@ -193,7 +193,9 @@ def run_Results_over(output):
     :type output: list
     """
     results_holder = []
+    pred_states_bylin = []
     for _, (tHMMobj, pred_states_by_lineage, LL) in enumerate(output):
         results_holder.append(Results(tHMMobj, pred_states_by_lineage, LL))
+        pred_states_bylin.append(pred_states_by_lineage)
 
-    return results_holder
+    return results_holder, pred_states_bylin
```
shak360, line 201: Ideally, this should be stored in the `results_holder` but I can add that on a new branch.

shak360: This would avoid having the `nothing` argument in the other functions.

Farnazmdi: Great, thanks!
I will merge this for now then.

PR #404, opened 2020-06-22
removes nothing argument and adds the predicted switched states as a key in the results_holder dictionary

shak360:


shak360:
@Farnazmdi can you check and approve if you feel like the stuff from the previous pull request is fixed here?

Farnazmdi:
It looks it covers it. Thanks.

PR #407, opened 2020-06-23
one figure maker function for the common figures

shak360:


shak360:
OK @jclagarde , looks like this is done. Look over it when you can. @Farnazmdi can you approve this after review. We just removed the second figure maker function for the exponential plots and combined it with the existing figure maker function for gamma. This is to fix #333 .

PR #414, opened 2020-06-26
figures arranged and figure 4 added

Farnazmdi:
This
image

converted to this:
image

I just grouped them and plotted a boxplot.
I used the accuracy score Dr. Meyer said, it does not look good. But this is the plot I was looking for. Do you think I should have done it another way to simulate @shak360 ?


lineage/figures/figure4.py
```diff
@@ -1,133 +1,108 @@
...
 import pandas as pd
+import scipy.stats as sp
+
 import seaborn as sns
 from .figureCommon import (
     getSetup,
     subplotLabel,
+    commonAnalyze,
     pi,
     T,
```
shak360, line 12: I'd recommend removing the import for `T`, just to make sure it's not conflicting with the T's in your for loop.

lineage/figures/figure4.py
```diff
@@ -1,133 +1,108 @@
...
+            tmp_lineage = LineageTree(pi, T, E2, 2**9-1)
+            good2go = lineage_good_to_analyze(tmp_lineage)
+
+        # how many cells are in state 1
+        cell_num_st1 = []
+        for cell in tmp_lineage.full_lineage:
+            if cell.state == 0:
+                cell_num_st1.append(1.0)
+
+        list_cell_num_st1.append(len(cell_num_st1))
```
shak360, line 64: The number of cells stored in state 1 is stored in https://github.com/meyer-lab/lineage-growth/blob/169ad6601af9fc8732f03bd7ea6b634caf92389a/lineage/figures/figureCommon.py#L116 so if you plan on using `commonAnalyze`, you can just call it with the argument `xtype="prop"`.

lineage/figures/figure4.py
```diff
@@ -1,133 +1,108 @@
...
+        population.append(tmp_lineage)
+
+        # Adding populations into a holder for analysing
+        list_of_populations.append(population)
+        list_of_fpi.append(pi)
+        list_of_fT.append(T)
+        list_of_fE.append(E2)
+
+    cell_nums, _, _, _, _, _, accuracy = commonAnalyze(list_of_populations, list_of_fT=list_of_fT)
```
shak360, line 73: Just add `xtype="prop"` and it'll turn `cell_nums` into proportion of cells in state 0, I think. The calculation is done in `Results`. See here: https://github.com/meyer-lab/lineage-growth/blob/169ad6601af9fc8732f03bd7ea6b634caf92389a/lineage/Analyze.py#L169 

`commonAnalyze` will return this for your populations, if you use that `xtype`. This works the same way as using`xtype="wass"`.

shak360:
You can also use figures 8 and 9 as references, as those figures do a similar analysis. I'd expect similar performance as what we see in figure 6, disregarding any small changes in accuracy due to lineage size.

Farnazmdi:
image
I did the comments and got this!

shak360:
OK, if it's not too much work, it might help to compare figures S08 and figure S09 with the new balanced_accuracy_score instead of the accuracy, and then make the comparison to what you've graphed here. @Farnazmdi

shak360:
> 
> 
> image
> I did the comments and got this! I mean it looks good.

Yea this looks a lot more reasonable and sorta what I would expect, poor performance at the sides and sometimes in the middle (because the transition matrix is nearly all 0.5).

Farnazmdi:
The performance is almost the same, this shows that the percentage goes from 0 to 100%, but what I showed earlier was just going from 0 to 50% basically half of this plot. But yea it is better to use this one. 
I will see what figureS07 would look like. (8 and 9 are wasserestein plots.)

shak360:
> 
> 
> The performance is almost the same, this shows that the percentage goes from 0 to 100%, but what I showed earlier was just going from 0 to 50% basically half of this plot. But yea it is better to use this one.
> I will see what figureS07 would look like. (8 and 9 are wasserestein plots.)

Sorry, yea figure S06 and S07.

Farnazmdi:
Here is what figure S07 look like with balanced accuracy score:
image

shak360:
It's better to compare it to figure S06, since that is uncensored.

Farnazmdi:
Yea ok, didn't notice that.
One question, why is the list of `pi`s given as argument? I mean shouldn't it be list of `T`s?
This is figure S06
image

shak360:
> 
> 
> Yea ok, didn't notice that.
> One question, why is the list of `pi`s given as argument? I mean shouldn't it be list of `T`s?
> This is figure S06
> image

No that is a list of fixed pis, so the model never has to estimate the pi. This is because the population are all single lineages so it doesn't make sense to estimate pi. Anytime there's an extra list_of_fXXXX, it means that that parameter is fixed for the mode.

Farnazmdi:
Okay, I think it looks better in scatter than boxplot. 
image

Farnazmdi:
Could you review please @shak360 ?

lineage/Analyze.py
```diff
@@ -169,6 +170,7 @@ def Results(tHMMobj, pred_states_by_lineage, LL):
     results_dict["state_proportions_0"] = results_dict["state_proportions"][0]
     results_dict["accuracy_before_switching"] = 100 * np.mean(ravel_pred_states == ravel_true_states)
     results_dict["accuracy_after_switching"] = 100 * np.mean(ravel_switched_pred_states == ravel_true_states)
+    results_dict["balanced_accuracy_score"] = 100 * balanced_accuracy_score(ravel_switched_pred_states, ravel_true_states)
```
shak360, line 173: Ah, sorry this is a really minor change but the true states should be the first argument and the estimated/predicted states should be the second argument. Feel free to do this on this branch or a new one.

lineage/figures/figure1.py
```diff
@@ -0,0 +1,138 @@
+""" This file contains functions for plotting different phenotypes in the manuscript. """
```
shak360, line 1: I thought figure 1 was supposed to be a diagram.

lineage/figures/figure2.py
```diff
@@ -66,7 +66,7 @@ def accuracy():
         list_of_fT.append(T)
         list_of_fE.append(E2)
 
-    total_cellnum, paramEst, accuracy_after_switching, transition_matrix_norm, pi_vector_norm, paramTrues = commonAnalyze(list_of_populations)
+    total_cellnum, paramEst, accuracy_after_switching, transition_matrix_norm, pi_vector_norm, paramTrues, _ = commonAnalyze(list_of_populations)
```
shak360, line 69: It might just be better to make the `accuracy_after_switching` the `balanced_accuracy_score` in general

lineage/figures/figure3.py
```diff
@@ -1,35 +1,31 @@
...
+import pandas as pd
+import scipy.stats as sp
+import seaborn as sns
 from .figureCommon import (
     getSetup,
     subplotLabel,
     commonAnalyze,
     pi,
-    T,
     E2,
```
shak360, line 13: This `E2` import isn't being used.

lineage/figures/figure3.py
```diff
@@ -41,140 +37,85 @@ def makeFigure():
...
+
+    wass, _, Accuracy, _, _, paramTrues, _ = commonAnalyze(list_of_populations, xtype="wass")
+    total = []
+    for i in range(4):
+        tmp1 = list(sp.gamma.rvs(a=paramTrues[i, 0, 3], loc=0.0,
+                                 scale=paramTrues[i, 0, 5], size=200))
+        total.append(tmp1)
+        tmp2 = list(sp.gamma.rvs(a=paramTrues[i, 1, 3], loc=0.0,
+                                 scale=paramTrues[i, 1, 5], size=200))
+        total.append(tmp2)
```
shak360, line 76: What is this for loop for?

lineage/figures/figureCommon.py
```diff
@@ -120,7 +120,7 @@ def commonAnalyze(list_of_populations, xtype="length", **kwargs):
     elif xtype == "bern":
         x = paramTrues[:, 0, 0]
 
-    return x, paramEst, dictOut["accuracy_after_switching"], dictOut["transition_matrix_norm"], dictOut["pi_vector_norm"], paramTrues
+    return x, paramEst, dictOut["accuracy_after_switching"], dictOut["transition_matrix_norm"], dictOut["pi_vector_norm"], paramTrues, dictOut["balanced_accuracy_score"]
```
shak360, line 123: Might just be easier to replace `dictOut["accuracy_after_switching"]` with `dictOut["balanced_accuracy_score"]` and keep the number of output the same

lineage/figures/figure3.py
Farnazmdi: Nice catch! I found the reason why the violin plots were not showing different distributions!

Farnazmdi:
All comments should be addressed with this commit.

lineage/figures/figure3B.py
```diff
@@ -65,6 +65,7 @@ def accuracy():
         list_of_fE.append(E)
 
     wass, _, Accuracy, _, _, paramTrues = commonAnalyze(list_of_populations, xtype="wass")
+    print(paramTrues.shape)
```
shak360, line 68: Remove this `print` statement if you're done debugging

lineage/figures/figureCommon.py
```diff
@@ -148,7 +148,7 @@ def overlayCartoon(figFile, cartoonFile, x, y, scalee=1, scale_x=1, scale_y=1):
     template.save(figFile)
 
 
-def figureMaker(ax, x, paramEst, accuracies, tr, pii, paramTrues, xlabel="Number of Cells"):
+def figureMaker(ax, x, paramEst, accuracies, tr, pii, paramTrues, balanced_acc, xlabel="Number of Cells"):
```
shak360, line 151: `balanced_acc` can be removed from this function list of arguments, or change `accuracies` to `balanced_accuracies` and remove `balanced_acc`. Make sure that in the rest of the function `balanced_accuracies` is replacing `accuracies`.

shak360: Pretty sure the figures won't build unless this change is made.

shak360: See here:
```
Traceback (most recent call last):
  File "./genFigures.py", line 20, in <module>
    ff = eval(nameOut + '.makeFigure()')
  File "<string>", line 1, in <module>
  File "/home/jenkins/workspace/meyer-lab_lineage-growth_fig4/lineage/figures/figureS02.py", line 34, in makeFigure
    figureMaker(ax, *accuracy())
TypeError: figureMaker() missing 1 required positional argument: 'balanced_acc'
```

shak360: The figures should be building before this is merged.

PR #424, opened 2020-06-30
Changes to how observations are imported 

shak360:


shak360:
@lkargi In this pull request, you can see what changes are about to go on master at the top right by looking at files changed, based on each commit. At the bottom of this page, you'll see some "checks" that need to pass before the changes are put on master. The first check is mainly for checking that all of our tests pass. Sometimes one or two tests will fail, but as long as the changes you made don't affect a large number of tests and cause them to fail, it should be fine. The other three checks are for merging (whether or not your code can be cleanly merged into master without conflicts), and for figure/manuscript building. These also check for linting, whether or not the code is properly written and follows the PEP8 style guide. Currently, the repo is unstable because pylint is identifying about 66 issues that are not within the PEP8 style guide. It'll report stable when that number goes to below 50. Linting can always be done on another branch so that's not a big concern. Mainly, these checks will also tell you if figures build based on your changes to the code, currently the code you've written affects none of the figures so this shouldn't matter.

PR #427, opened 2020-07-05
AIC

shak360:


shak360:
@lkargi When you're done, we have to make sure that this is mergeable with `master`. This means two things, we have to resolve conflicts given a possible merge with `master`, and then merge with `master`. 

I find it easier to use this tool here (the `Resolve conflicts` button below). What you have to do is pick the code that you want to keep. It'll delineate conflicts between sections of `<<<<<<< luka-AIC` {the code you're suggesting to add} `========` {the code that is currently on `master`} `<<<<<<< master`. Usually, what you'll want to do is keep the code you're suggesting to add, and delete the stuff that is currently on `master`. There are, however, cases where you want to keep the stuff on master. So pick the section of code you want to keep, delete the other, and then delete the marking code used to delineate the two sections (the `<<<<` and the `====`).

Once you do that, you have to merge with `master`. You can do that here, using the button that it gives you, or using the command line using the following commands:

```
git status
git add -A
git commit -m "final"
git push
git checkout master
git pull
git checkout luka-AIC
git merge master
(Ctrl-X)
```

On the top right, add me as a reviewer when you think it's ready.

PR #430, opened 2020-07-07
Fig6

Farnazmdi:
Added fig6.
Included figures to the result part.
removed fig 0 and moved the lineage tree plots to figure 6.


Farnazmdi:
I can't find what is that it is failing the build but should be fine, Jenkins passed.

I made changes to the figures, but I had to remove figureS25 and S24 from the `make` file to run fast, they were taking forever to run.

Farnazmdi:
When do you think the AIC will be merged? @shak360

shak360:
> 
> 
> I can't find what is that it is failing the build but should be fine, Jenkins passed.
> 
> I made changes to the figures, but I had to remove figureS25 and S24 from the `make` file to run fast, they were taking forever to run.

They're taking forever to run because, the lineages they're trying to create and analyze are too small or aren't heterogeneous enough for it to pass `good2go`. The only fix would be to change the parameters to be lower (in terms of the gamma distributed times) or increase the experiment time. I wouldn't worry about it though.

Farnazmdi:
The figures are looking good now, I guess we can modify the parameters to create those figures later.

shak360:
The figures look good. Is there anything, in particular, (python code logic or otherwise) you want me to take a closer look at? Otherwise, feel free to merge.

shak360:
> 
> 
> When do you think the AIC will be merged? @shak360

image

Here's what the figure looks like now. Luka's making some minor changes and it should be done.

Farnazmdi:
I don't see any grid-style changes. I will merge this and let's take care of that in another branch, but I didn't mean to change any grid-style.

PR #434, opened 2020-07-09
Some figure cleaning (2,3,4)

shak360:


genFigures.py
```diff
@@ -98,8 +95,14 @@
...
+#                        f'{cartoon_dir}/figure2.svg', 120, 0, scalee=0.675)
         
-    if sys.argv[1] == '2':
-        # Overlay Figure 2 cartoon
-        overlayCartoon(fdir + 'figure2.svg',
-                       f'{cartoon_dir}/figure2.svg', 0, 0, scalee=0.3)
+#     if sys.argv[1] == 'S52':
+#         # Overlay Figure 2 cartoon
+#         overlayCartoon(fdir + 'figureS52.svg',
+#                        f'{cartoon_dir}/figure2.svg', 120, 0, scalee=0.675)
```
shak360, line 108: The cell lineage cartoon with the observations will go here.

genFigures.py
```diff
@@ -83,9 +80,9 @@
                        f'{cartoon_dir}/figureS05.svg', 65, 0, scalee=0.3)
 
     if sys.argv[1] == '4':
-        # Overlay Figure 3b cartoon
+        # Overlay Figure 4 cartoon
         overlayCartoon(fdir + 'figure4.svg',
-                       f'{cartoon_dir}/figure3B.svg', 40, 30, scalee=0.24)
+                       f'{cartoon_dir}/figure4.svg', 0, 30, scalee=0.235)
```
shak360, line 85: Renaming the 3b cartoon to 4.

shak360:
image

Figure 2 (as it is in the manuscript). Uncensored lineage statistics and the estimation of them.

shak360:
image

Figure 2, but is censored (I just put this in the supplementary figures, doesn't necessarily have to be in the manuscript). I just wanted to see what it looks like. We should keep in mind that I've turned the random seed off.

shak360:
image

Figure 3 as it is in the manuscript. Estimation results over varying population sizes (the number of lineages is increased). This is for censored lineages.

shak360:
image

Figure 3 but the uncensored version. Again, it's a supplementary figure, but doesn't necessarily mean that it has to be in the manuscript or anything.

shak360:
image


Figure 4, as it is in the manuscript. These are uncensored lineages over increasing state distance. 4 lineages in particular are picked to have their G1 lifetime observation plotted to show what state distance (operationalized by the Wasserstein distance) means. The lineplot of accuracy is done the same way as it is in figure 3, but with the x-axis being the Wasserstein distance.

shak360:
These are all screenshots, by the way, so they look better on Jenkins and don't have things cut off.

shak360:
@Farnazmdi feel free to go in and make changes if you want. I'll work on figures 5 6 7 on a new branch (5 being the state representation, 6 being AIC, and 7 being the censored lineage figure).

shak360:
I increased the number of data points to 200, so the build will take longer now.

Farnazmdi:
They look very good! @shak360 
In figure4, I guess we should decrease the gap between the cartoon and the second subplot.

In general, the `errorfill` looks better than the boxplots, the problem is when the data is grouped, it is making discrete chunks looks sharp, I mean if we had not grouped a couple of data points together, then the ribbon should have been more smooth. I don't have a solution immediately, but I think it is something we can figure out by smoothing or something...

In figure 2, the second row parameters (corresponding to G2, I guess) should change because one of them looks like doesn't exist.

Other than these, they look very nice! Thanks guys.

Farnazmdi:
image
I think to make this more readable the grid size needs to be adjusted.

shak360:
image

Farnazmdi:
It looks like we cannot do scatter inside the errorfil(). So maybe let's only keep errorfill for accuracy and use errorbar or something else for parameter estimations.

PR #452, opened 2020-07-15
Figure grid modifications

shak360:
I think as a way to do the grid sizes, what I've found is that the following schema works:

For 4 rows by 4 column figures (where the individual figures are not that detailed), the width and height should be 10 inches and 10 inches respectively.

`ax, f = getSetup((10, 10), (4, 4))`

For these under-detailed figures, you can then scale the above depending on how many figures you want. If you want 2 rows by 2 columns, then the grid size in inches should be 5 in by 5 in.

`ax, f = getSetup((5, 5), (2, 2))`

If you want 3 rows and 1 column, then the grid size in inches should be, 2.5 in by 7.5 in. (Note how the order is different: the grid size in inches is width by height, but the grid shape is in rows by columns.)

`ax, f = getSetup((2.5, 7.5), (3, 1))`

===

If the figures are detailed, then a 3 rows by 3 columns figure, should also use 10 in by 10 in.

`ax, f = getSetup((10, 10), (3, 3))`

For these detailed figures, you can then scale the above depending on how many figures you want. If you want 2 rows by 2 columns, then the grid size in inches should be 5 in by 5 in.

`ax, f = getSetup((6.666, 6.666), (2, 2))`

If you want 1 row and 3 columns, then the grid size in inches should be, 3.333 in by 10 in.

`ax, f = getSetup((10, 3.333), (1, 3))`

===

In the rare case where you don't want a figure to be square, then it is recommended to keep the grid size in inches the same as usual, but alter the number of rows and column. For example, if we want the figures to extend horizontally, what we essentially want is to fit a 2 row by 1 column figure into a 2 row by 2 column figure. Because a 2 row by 2 column figure will take usually a 5 in by 5 in grid size, we will utilize that grid size, but use a 2 row by 1 grid shape.

`ax, f = getSetup((5, 5), (2, 1)) # each figure will take twice its normal size horizontally`

===

To be consistent, I hope we can follow this across the rest of the figures (at least the main figures and not necessarily for the supplementary ones).


shak360:
Figure 7 is taking an unusually long time to build.

PR #464, opened 2020-07-22
Fig6 b

Farnazmdi:
I added a figure file (not in the makefile yet) for aic of the experimental data. It doesn't run yet, though.
Collected lineages with the same condition from different experiments or replicates (not all of them) and put it in a file named `Lineage_collections`. We can import them by name, like `Lapatinib_Control`.
Added a function in this file so it removed lineages with their size <= 7.
I thought doing this might help avoid throwing that error, but it did not. 
We need to take a deeper look at that and what's the problem.
For now, I think it is enough code to be merged.


lineage/figures/figure6B.py
```diff
@@ -0,0 +1,65 @@
...
+    # Setting up state distributions and E
+    Sone = StateDistribution(0.99, 0.9, 10, 2, 10, 2)
+    Stwo = StateDistribution(0.9, 0.9, 20, 3, 20, 3)
+    Sthree = StateDistribution(0.85, 0.9, 30, 4, 30, 4)
+    Sfour = StateDistribution(0.8, 0.9, 40, 5, 40, 5)
+    Eone = [Sone, Sone]
+    Etwo = [Sone, Stwo]
+    Ethree = [Sone, Stwo, Sthree]
+    Efour = [Sone, Stwo, Sthree, Sfour]
+    E = [Eone, Etwo, Ethree, Efour]
```
shak360, line 34: I don't think these lines of code are needed. Should probably be removed.

shak360: Sorry line 23 is needed. I don't think the rest are.

lineage/figures/figure6B.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+desired_num_states = np.arange(1, 8)
+
+
+def makeFigure():
+    """
+    Makes figure 10.
+    """
+    ax, f = getSetup((13.333, 3.333), (1, 4))
```
shak360, line 22: This should just be one figure. The x-axis is the number of states to be picked, and the y-axis is the AIC value as the model tries to find that many states. Why are four subplots needed?

shak360: Are there 4 data sets?

shak360: Never mind, I just noticed that you want this to be on master before investigating further.

lineage/figures/figure6B.py
shak360: Never mind, I just noticed you wanted this on master before investigating further.

shak360:
Feel free to merge whenever @Farnazmdi .

PR #473, opened 2020-07-27
Added Content and Caption to AIC figure

lkargi:


shak360:
@lkargi don't worry about errors with codecov. I'm not entirely too sure what determines the passing/failure for them, especially when they pass on some pull requests and fail for others.

PR #480, opened 2020-07-29
figure 9 for experimental data AIC

Farnazmdi:
These are my observations:
1. the lesser the number of lineages in the population used as data, the lesser the fit fails! (I tried 10 lineages and it failed almost 1/3 of the time.)
2. I wanted to try bootstrapping (like 10 lineages at a time) but the problem is at least one of them fails in a for loop of range(5). The other problem is, the states won't stay the same and change identity randomly, so if we append them, we won't know which one is actually state 0 or 1 to be able to compare or see some kind of distribution for the parameters.
3. The answer of `Analyze` and of `.fit()` are almost the same (well, should be!).


aarmey:
@Farnazmdi (1) what does fail mean? (2) Are there assertions in the state distribution class to ensure it's not passing back absurd values? There should be—if something is nan or inf we should just immediately crash.

aarmey:
(Feel free to ignore this until next week. Realize you're taking the week off.)

Farnazmdi:
Sorry, didn't mean to ignore, I didn't have my laptop with me, it was hard to change things by phone.

Here is the assertion failure by which I meant it failed:
The assertion is in `UpwardRecursion` in the normalizing factor vector which has become zero in this case. 
image

Farnazmdi:
My opinion about lower accuracy in the phase state model is in the likelihood calculation. In the state distribution of phase model, we have 4 components multiplied together (bern_llG1 * bern_llG2 * gamma_llG1 * gamma_llG2) but in the simple gamma model, we only have 2 (bern_ll * gamma_ll).

And in case of assertion failure, could it be that when the 4 likelihood components are multiplied together, create a zero that propagates through and we see it in Upward recursion?

aarmey:
The idea would be to avoid any bad outputs. I'm not sure that 0 in one state is bad, but 0 for all states would be.

Farnazmdi:
If the zeros in NF is actually coming from the likelihood, it probably causes problems anyway, because I think NF is in some denominator in the process (to calculate betas).

Farnazmdi:
I may have found the problem with fitting the experimental data.
Would it be alright if we remove those cells that are present at the beginning of the experiment and their G1 is left-censored?
Here is an example of a lineage that if it is in the population that we are trying to fit the model to, the normalizing factor throws the assertion error.
image

Farnazmdi:
I am waiting for Shakthi's reply but one other possibility might be the way that the last two observations are being assigned:
image
They probably should be only zero or one, but we have `nan` here...

aarmey:
Censored phases should be treated as censored, not removed.

aarmey:
The rvs function is only used when generating data, not fitting.

Farnazmdi:
image
this is the AIC for the control lineage. I think it makes sense to have 1 state.

aarmey:
How many cells is that then?

Farnazmdi:
71 cells. These are lineages that cell number in each of them is greater than 7.

Farnazmdi:
The problem with that lineage (and probably those others that would result in the failure of that assertion) was that they contained a cell that its G1 or G2 duration was 0 and that 0 is not defined in Gamma distribution.

aarmey:
Was it at the beginning or end of the experiment? Those sound like cells where those phases we not observed (and so should be marked as such).

Farnazmdi:
Most likely at the end. Sure, will do.

Farnazmdi:
I am waiting for Luka's branch of importing to merge this.

Farnazmdi:
It doesn't look like the model finds more than 1 state in the real data, at least in the low concentrations:
image

aarmey:
Think you have to look at high concentrations, too. We’d expect the distributions to be more different there.

Farnazmdi:
The problem with those trials is that cell numbers are really low, the gamma estimator throws an error.
image
I think in this set of data, in each experiment, cells are treated with a drug at t=0. I will try to find something that works, but I guess a better type of data for this model would be the one that cells grow to some time point, and then the drug is introduced, so cell numbers wouldn't be so low from the beginning.

aarmey:
Hmm... are you sure this is from low cell numbers... "divide by zero encountered in **log**" makes me think it might be the data itself being handled strangely...

Farnazmdi:
We trusted Luka's changes in handling data importing. You are right, I filtered lineages smaller than 9 cells, and still, I get this warning. I will go through the file again to see if I can find anything that has slipped out.

aarmey:
Well, if I had to guess, it's not an issue with the raw data or import. I would first look to see if censoring/non-observed isn't being handled properly.

Farnazmdi:
Figure 8's structure is ready, I will merge this if it is alright with you, then I will work on fixing the transition matrix and other parameters on another branch. If that works, it will be only 1 line change in this file. @aarmey 
image

genFigures.py
```diff
@@ -4,6 +4,8 @@
 import time
 import matplotlib
 import numpy as np
+import random
```
aarmey, line 7: Uncomfortable doing this.

lineage/BaumWelch.py
```diff
@@ -16,6 +18,7 @@
     sum_nonleaf_gammas,
 )
 
+logging.basicConfig(filename='example.log',level=logging.DEBUG)
```
aarmey, line 21: This should be set in genFigures.

lineage/BaumWelch.py
```diff
@@ -97,11 +101,14 @@ def do_M_T_step(tHMMobj, MSD, betas, gammas):
         gamma_array = gammas[num]
 
         # local T estimate
-        numer_estimate += get_all_zetas(lineageObj, betas[num], MSD[num], gamma_array, tHMMobj.estimate.T)
-        denom_estimate += sum_nonleaf_gammas(lineageObj, gamma_array)
+        numer_estimate = get_all_zetas(lineageObj, betas[num], MSD[num], gamma_array, tHMMobj.estimate.T)
```
aarmey, line 104: Are you sure this is right? You're only going to end up estimating based on the last iteration of the loop...

lineage/DownwardRecursion.py
```diff
@@ -1,12 +1,16 @@
 """ File holds the code for the downward recursion. """
 
+import math
 import numpy as np
+import logging
 from .UpwardRecursion import beta_parent_child_func
 
+logging.basicConfig(filename='example.log',level=logging.DEBUG)
```
aarmey, line 8: Remove.

lineage/LineageInputOutput.py
```diff
@@ -193,15 +191,15 @@ def tryRecursion(pColumn, lower, upper, parentCell, currentLineage, data, divisi
...
-    if exp_time == -1 and not math.isnan(data[parentPos][pColumn + 2]) and daughterCell.left is None and daughterCell.right is None:
-        exp_time = data[parentPos][pColumn + 2]
+    if exp_time == -1 and not math.isnan(data[parentPos][pColumn+2]) and daughterCell.left == None and daughterCell.right == None:
+        exp_time = data[parentPos][pColumn+2]
         if global_exp_time[0] != -1:
             assert exp_time == global_exp_time[0], f"Exp_time discrepancy in file {exp_time} and {global_exp_time[0]}"
     if global_exp_time[0] == -1 and exp_time != -1:
         global_exp_time[0] = exp_time
-
+
```
aarmey, line 202: Please lint with autopep8 or black.

lineage/UpwardRecursion.py
```diff
@@ -1,6 +1,8 @@
 """This file contains the methods that completes the downward recursion and evaulates the beta values."""
-
+import math
 import numpy as np
+import logging
+logging.basicConfig(filename='example.log',level=logging.DEBUG)
```
aarmey, line 5: Remove.

lineage/tHMM.py
```diff
@@ -1,12 +1,14 @@
...
+import math
 import numpy as np
+import logging
 import scipy.stats as sp
 
 from .UpwardRecursion import get_Emission_Likelihoods
 from .BaumWelch import do_E_step, calculate_log_likelihood, do_M_step, do_M_E_step
 from .Viterbi import get_leaf_deltas, get_nonleaf_deltas, Viterbi
 
+logging.basicConfig(filename='example.log',level=logging.DEBUG)
```
aarmey, line 11: Remove.

lineage/BaumWelch.py
Farnazmdi: Sorry, I was messing with it and forgot to return it back to the original version and accidentally committed.

Farnazmdi:
> A few edits. Also, what are the changes here. New asserts, file loading, and added manuscript text?

Yes, exactly. The text for the intro is in another branch though.

Farnazmdi:
Actually, now that I think about it, none of the loggings and assertions really helped. I will move the two files that are the important changes in this PR to a new PR and work on fixing the parameters of the model, if it is alright @aarmey

aarmey:
Sounds good.

PR #490, opened 2020-08-06
Figure 1 update

Farnazmdi:
Figure 1 updated:
image

I tried to bring the subfigure label (b) on the middle up part of the figure, but haven't been successful.. here is my line:

`   ax[0].text(-0.2, 1.25, ascii_lowercase[0], transform=ax[0].transAxes, fontsize=16, fontweight="bold", va="top")
    ax[1].text(0.6, 1.25, ascii_lowercase[1], transform=ax[1].transAxes, fontsize=16, fontweight="bold", va="top")`
First line for (a) and second line for (b)
but it doesn't appear.


Farnazmdi:
@shak360 do have any suggestions for bringing the subplot labels?

shak360:
Are those lineages from the data. Pretty cool regardless, nice work. Also, you can use different colors other than olive and red haha, I was just messing around when I chose those colors, I'm sure there's a better color than olive.

@Farnazmdi 

For the subplot labels, because the svgs are pasted onto the generated matplotlib figure _after_ the labels are made, it's pretty easy for the label to get hidden under the lineage svgs. What you could try to do, is somehow set an option for the lineage tree svgs to be made transparently. 

So: try adding `transparent=True` to the line in `plotTree.py`.

```
pylab.savefig(path, format="svg", bbox_inches="tight", dpi=400)
```

Farnazmdi:
They are from the data yeah. Thanks!

The options for color are limited for `draw()` in Phylo, so these (and two other colors that show censoring) were the best colors I could find. But yea if there is a better option, I am all ears :)

Awesome! The transparent worked! Thanks

Farnazmdi:
image

PR #493, opened 2020-08-06
organize manuscript and figures

Farnazmdi:
Organized the supplementary figures.
Removed some and changed the order of others.
All the main and supp. figures are being built now. (including figure3, with a new cartoon)
wrote methods for data and dealing with censorship. (might need a little more than just an edit)
All parts of the manuscript should be a bit more ready.
JC and Luka are working on figures 6 and 7.
(Sorry for many changes in 1 PR...)

** I thought maybe for figures 5 and 6 and S07 (Wasserstein and % in state 0) in the main manuscript to create the lineages by increasing both depth and breadth, but it looked complicated to do because one of them was creating lineages by looping over the `T`s and the other looping over the `E`s. If you agree with this change, maybe @shak360  you can help with the changes?


shak360:
I think it might be worth to reduce the number of lineages figure 3 is producing or remove it form the makefile @Farnazmdi otherwise the build takes a while. Still unsure of why it's doing this.

shak360:
> I think it might be worth to reduce the number of lineages figure 3 is producing or remove it form the makefile @Farnazmdi otherwise the build takes a while. Still unsure of why it's doing this.

Or I might be wrong, seems like the builds are taking a normal amount of time...

shak360:
@Farnazmdi Is there a particular file or set of files you'd want me to review?

Farnazmdi:
Yeah, do you agree to create populations for figure 5 (% in state0) and 6 (wasserestein), and S07 with increasing both number of lineages and number if cells? @shak360 I think it is closer to real and we could avoid those few times that the build fails due to these figures.

shak360:
@Farnazmdi so if those figures also have increasing population size, what is represented on the x-axis, if, say fo figure 5, both the percentage of cells in state 0 and the number of cells is increasing? That was why it made sense for me to have a fixed size of the lineages for those figures. Let me know if I'm misunderstanding something

Farnazmdi:
Oh no you are right. I meant the fixed population would be composed of more than 1 lineage. Does this make sense?

shak360:
> 
> 
> Oh no you are right @shak360 . I meant the fixed population would be composed of more than 1 lineage. Does this make sense?

Got it, Yea I agree with the changes then, Great work : )

Farnazmdi:
I haven't made those changes yet. About to do in another branch.

PR #496, opened 2020-08-09
figures 5, 6, S07

Farnazmdi:
Figures 5 and 6 and 7 have multiple lineages now (multiple starting cells at the beginning), instead of a single lineage.
S07 looked a lot better, but not sure about figure 6. Figure 5 was good already. 
The Viterbi test is fixed.


Farnazmdi:
I think it is worth having only an accuracy plot against cell number for 3-state model, too. It looks like the test in tHMM that tests the performance of a 3-state lineage, fails if the cells are less than ~2e11 or 2e12.

Farnazmdi:
@shak360 I don't think they need deep review, all are what we have discussed in the slack with Dr. Meyer plus some linting. If you could, please take a look at the Viterbi test.

Farnazmdi:
The accuracy still remains an issue. The linting still fails (but I'm not sure how much more I can do the linting). tHMM tests fail less, but still fail.

lineage/tests/test_Viterbi.py
```diff
@@ -18,11 +17,16 @@ def test_vt(self):
...
+        log_scores_flipped = tHMMobj.log_score(pred_states_flipped)
 
         for _ in range(10):
             rand = randint(0, 2, (2 ** 9) - 1)
             # generate a sequence of integers between 0 (inclusive)
             # and 2 (exclusive)
             random_log_scores = tHMMobj.log_score([rand])
-            self.assertTrue(random_log_scores[0] <= true_log_scores[0])
+            model_log_score = max(log_scores_orig[0], log_scores_flipped[0])
+            self.assertTrue(random_log_scores[0] <= model_log_score)
```
shak360, line 32: Yep, this is exactly the required change 👍

PR #502, opened 2020-08-13
Fix import

lkargi:


lkargi:
Should be good to merge now.

lineage/LineageInputOutput.py
```diff
@@ -1,11 +1,13 @@
...
 import math
 import pandas as pd
 from .CellVar import CellVar as c
+import numpy as np
 
 
-def import_Heiser(path, exp_time=192):
+def import_Heiser(path):
+    global global_exp_time
+    global_exp_time = -1
```
shak360, line 10: I don't think you need the `global` keyword being used here in line 9. Defining it in line 10 the way you do should be fine. If you want this variable to be exposed to everything in the file, then just write line 10 in line 7 outside of the function.

lkargi: It throws an error when I don't declare it global first though, because I'm using it the the recursive calls too though.  I'm not sure I really understand what the global keyword does exactly. I should only be removing about 5 excel files and adding back new ones, the rest of them I just added exp_time.

lkargi: The way I understand it is that global tells the function to use the global x. So if I don't declare it global at first, the global version of x won't be initialized.  Then when I try to use it in the recursive function without setting it the global version of x hasn't been initialized yet so it throws an error.  That might be totally wrong though I'm not sure.

shak360: Do you get this error even when setting this variable outside of `import_Heiser`? If so, then I'm comfortable with keeping it as is and figuring it out later.

OK, and yea I just wanted to make sure that you were aware that all these Excel files were being changed. About **200** of them are being changed with this pull request.

lkargi: The issue I have with setting it outside is that then it won't reset with every import_heiser call. So if you run the import function 2 times after import, global exp_time will start at -1 the first call, and then initialize to some value.  then the second time it will start at that new value because it's global, and without the code in the import_heiser call it won't reset to -1.  I guess Ideally it would be "global" to tryRecursion and local to import_heiser but I don't know how I could do that.

shak360: Could this work:

```python
global_exp_time = -1
def import_Heiser(path):
    global_exp_time = -1
    ...
```

lkargi: there the global_exp_time  would not be a local variable in import_heiser so updating it there doesn't affect the global one. I think I could get around this global stuff by putting global_exp_time into a list so its passed by reference, if you want to try to avoid using global.

shak360: I think its OK as is, maybe include some comments as to why `global` keyword was required, and that should be fine

shak360: Oh and you're meaning to change all these excel files, right, all 200ish of them?

lkargi: `def import_Heiser(path):
    global_exp_time = [-1]`
`def tryRecursion(pColumn, lower, upper, parentCell, currentLineage, data, divisionTime, exp_time, global_exp_time):`
How about this? Its still a bit clunky but I think its better than using global.

shak360: That looks good :)

lkargi: yeah and for the excel files I added a column with the exp_time to all of them

shak360: Cool, merge whenever you're comfortable

lkargi: I can't still cause I'm not authorized

PR #505, opened 2020-08-18
Iss476

lkargi:
@aarmey This is the branch trying to implement vectorization for gamma_sf in the estimator


aarmey:
Hmm... I think you have the function in vectorized form, but still need to call it in vectorized form. What I mean by this is instead of calling it repeatedly for each cell, you pass in all the cells at once. What makes me think this isn't the case is there are still the if statements for checking for censored/unobserved cases, and these are not vectorized.

Making this change will be very tricky, though. I'd suggest leaving this open, and then I'd be interested to try and get it working one day.

lkargi:
Yeah I hadn't really worked on implementing it into the estimator yet, just changed the functions.  The other issue was that it just runs slower than the the scipy.stats.gamma.sf function when there are more than ~10 observations.  So I'm not sure how useful this would even be compared to the scipy version.  I do want to push the new sf with njit though, since that runs faster with just 1 observation.

aarmey:
That's understandable—vectorization provides a very small amount of overhead, and so only speeds up the calculation when called with many elements at once.

PR #521, opened 2020-08-26
tests for 3-cell lineages

Farnazmdi:
2 tests for lineages with 3 cells added.
1 for simple gamma distribution, and the other test for phase-specific gamma.
The problem of nan propagation is observed even when we create synthetic cells.
They don't fail *always*, both of the tests sometimes fail. The one for phase-specific fails much more often. But I saw a case when both failed, and a case when both passed and a case that just the phase-specific one failed.


aarmey:
Actually, this only touches testing which passed, so I'll merge.

PR #523, opened 2020-08-26
manuscript edits and figure 9 correction.

Farnazmdi:
@aarmey I corrected the number of parameters according to this paper: [https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.102.7556&rep=rep1&type=pdf](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.102.7556&rep=rep1&type=pdf)
And now it is estimating more states: (the majority is 1 yet, though)
image

This is the number of parameters, assuming we have `k` states. originally we have 6 parameters (2 Bernoulli, 2 pairs of gamma shape and scale). If we keep shape parameters of G1 and G2 constant, we have 4 parameters per state.
transition matrix: `k * (k - 1)`
initial prob. matrix: `(k - 1)`
distribution parameters: `k * 6`, but if we have constant shapes: `k * 4`.

I wrote a paragraph explaining that in the results, in the very last figure, the very last paragraph.
(I made a few more changes to the manuscript that can be ignored for now.)


aarmey:
Excellent! Just realized something @Farnazmdi... how are there multiple lines on these plots if there's only one set of data?

Farnazmdi:
I guess each of them is a lineage.

aarmey:
Hmm... we should be looking at the overall AIC across lineages. That is, the log-likelihood can be summed across lineages (then you do the AIC correction).

Farnazmdi:
I think it worked.

aarmey:
So this looks right for the real data now, but the simulated case should have multiple trials. As that data is generated randomly, we want to show multiple trials of the full process to see the distribution of outcomes.

lineage/figures/figure8.py
```diff
@@ -1,55 +1,121 @@
 """
-File: figure7.py
-Purpose: Generates figure 7.
+File: figureS11.py
+Purpose: Generates figure S11.
```
aarmey, line 3: Isn't this figure 8?

lineage/figures/figure8.py
```diff
@@ -1,55 +1,121 @@
...
-from .figureS11 import run_AIC, figure_maker
+from .figureCommon import getSetup, lineage_good_to_analyze, subplotLabel
 
+# TODO
+# Calculate A-AIC where you removed the count of elements in T and pi that are zero, and add for those that are 1.
 
 desired_num_states = np.arange(1, 8)
 
+# Setting up state distributions and E
+Sone = StateDistribution(0.99, 0.9, 10, 2, 10, 2)
```
aarmey, line 24: Don't we have a common set of these in figureCommon?

lineage/figures/figure8.py
```diff
@@ -1,55 +1,121 @@
...
 
     # Finding proper ylim range for all 4 uncensored graphs and rounding up
-    upper_ylim_uncensored = int(1 + max(np.max(np.ptp(AIC[0], axis=0)), np.max(np.ptp(
-        AIC[1], axis=0)), np.max(np.ptp(AIC[2], axis=0)), np.max(np.ptp(AIC[3], axis=0))) / 25.0) * 25
-
-    # Finding proper ylim range for all 4 censored graphs and rounding up
-    upper_ylim_censored = int(1 + max(np.max(np.ptp(AIC[4], axis=0)), np.max(np.ptp(
-        AIC[5], axis=0)), np.max(np.ptp(AIC[6], axis=0)), np.max(np.ptp(AIC[7], axis=0))) / 25.0) * 25
-
+    upper_ylim_uncensored = int(1 + max(np.max(np.ptp(AIC[0], axis=0)),
```
aarmey, line 44: What is this trying to calculate? I have a feeling it can be much simpler...

lineage/figures/figure8.py
Farnazmdi: corrected. I replaced all of the figure S11 to figure 8 (everything was the same.)

lineage/figures/figure8.py
Farnazmdi: we only have 2 stateDistributions defined in figureCommon, here we need 4, and these parameters have to be well-chosen so that they are different from one another.

lineage/figures/figure8.py
Farnazmdi: It is trying to find a common maximum for y_ticks between all 4 plots in a row.

PR #526, opened 2020-08-27
Merging the cell cycle phase StateDistribution

aarmey:
The cell cycle phase StateDistribution is essentially the Gamma one stacked back to back. The only exception is that G2 is sometimes completely unobserved. Rather than redefining a bunch of code, we should simply be re-using the Gamma SD in `rvs()`, `pdf()`, and `estimator()`. The estimator will be a little more difficult because of how the observations are unpacked. I think they'll have to be unpacked, separated, then repacked into the tuple form. Alternatively, the code after unpacking within the Gamma estimator could be made into a common function.

This will (1) simplify the code significantly, and (2) likely uncover what the issue is when handling separate cell cycle phases.


aarmey:
@Farnazmdi could you make an attempt at finishing this?

Farnazmdi:
Of course. @aarmey

Farnazmdi:
@aarmey sorry, how can I build this to see if there are errors? It is not being built.

aarmey:
Check on Jenkins. The branch should be building. Not sure why it's not linked to Github.

Farnazmdi:
Oh, I see that now. Thanks.

Farnazmdi:
@aarmey should be done.
The test that fails (I checked for 2, 3 times) is the estimation of one of the parameters:
image
I don't know why.

aarmey:
Looking into it...

aarmey:
Look at last commit. The ordering of the parameters is confusing, and two of the indices were swapped.

aarmey:
Great work! If this builds and the figures look right then this is ready to merge.

Farnazmdi:
oh my god sorry... the order was wrong..
Thank you.

Issue #524, opened 2020-08-27
Lineages should be directly added to plots, not saved as SVG


aarmey:
See information on how to do so here: https://stackoverflow.com/questions/17458580/embedding-small-plots-inside-subplots-in-matplotlib

Farnazmdi:
I did it for figure 1, but I can't find a way to rotate the tree plots by 90 degrees like what we have right now. I had done it with the overlay_cartoon function. Any function I found that could rotate, needed an image. Would it be alright to keep the tree plots (cartoons) for figure 4 as it is right now? @aarmey

PR #536, opened 2020-08-30
embed lineage plots in the figures

Farnazmdi:
Figure 1 and figure 4 were modified based on issue #524.
The difference is that I couldn't rotate them. If you think they might not be suitable in terms of their height and length in the paper, please let me know @aarmey.


aarmey:
How about something like this? https://stackoverflow.com/questions/22540449/how-can-i-rotate-a-matplotlib-plot-through-90-degrees

aarmey:
Actually, I like them better in their orientation here. I think it makes sense for the x-axis to be time.

Farnazmdi:
image
They show up for me, are you referring to something else? (I only kept 2 trees of censored and 2 of uncensored to save up some space, they used to be 3.)

aarmey:
You're right. Can't find the missing one now...

PR #548, opened 2020-08-31
Test state distance

aarmey:


aarmey:
Still need to revert the numba removal, but make the overall pdf calculation one larger numba function.

PR #555, opened 2020-09-02
aic edits

Farnazmdi:
In this PR I removed the number of parameters as an argument from the user, and placed the constant parameters that are passed in the `tHMM` instead of `fit()`.
It will count the number of parameters based on the `constant_params` given to it.
increased the number of states for figure 9 from 4 to 6.


aarmey (changes requested):
I'm not sure why this has to be passed through all these functions. Can't this just be an argument within the emissions class (`E`) when you create it?

Farnazmdi:
@aarmey I hadn't thought of passing that through `E`. Will try it right away.

Farnazmdi:
I tried some changes, but couldn't find a good way of changing the `E`. We only define a dummy `StateDistribution` for the data that we import. I tried assigning the constant values after fitting, through `tHMMob.estimate.E[state].aprams[2, 4] = constant` but I think this is after fitting and it doesn't help the way we wanted it.

aarmey:
Couldn't the StateDistribution have an argument within `__init__()` like `const`, that defaults to `None`, but that you can set to define the constraint?

Farnazmdi:
I thought it will get overwritten. I will try it.

aarmey:
Hmm... I think the emissions are updated in-place. If not we can fix that.

PR #561, opened 2020-09-03
Adjusted the figure tagging in the Results section of the manuscript

jclagarde:


aarmey (approved):

Farnazmdi:
@jclagarde  one suggestion. I see sometimes you use `Fig. something` sometimes `Figure something`. I am not sure which one we prefer, but I think Dr. Meyer used `Fig. something`.

jclagarde:
> @jclagarde one suggestion. I see sometimes you use `Fig. something` sometimes `Figure something`. I am not sure which one we prefer, but I think Dr. Meyer used `Fig. something`.

@Farnazmdi ohh I see. Okay, I will edit it to be consistent to Fig. something !

PR #563, opened 2020-09-03
fig6

Farnazmdi:
In this PR:
In figure 6 the KMeans added, and the Wasserstein distance is added to the distribution plot.


Farnazmdi:
image
Here is the updated figure 6.
(Kmeans doesn't do this bad every time I run it)

aarmey:
The baseline should be able to do better than 0.5 when the states are really different. What about using just above or below the median as the baseline? That should get an accuracy > 0.5 when they are quite different...

Farnazmdi:
I had changed the parameter set, so I changed it back and it looks reasonable now.
Sorry I didn't understand what you meant by using the median as a baseline.

aarmey (changes requested):
This looks much better. The only change I see is sometimes kmeans gets the labels backwards, so the accuracy is 0.2 when it should be 0.8. I think you can simply say that if the accuracy is less than 0.5, you'll correct it to be 1 - x.

Farnazmdi:
image
Corrected, thank you.

PR #565, opened 2020-09-04
Aic3

aarmey:


aarmey:
@Farnazmdi this removes all the const passing—you can now just set it in the emissions definition. I think this branch just needs:

- [ ] figure9.py updated
- [ ] The estimator with const defined is not handling censoring correctly—it needs to run the censored estimator fitting with a restricted bound on the shape
- [ ] Rather than looking at the params vector, StateDistribution should have a method to provide the degrees of freedom. That way the StateDistribution can check for const internally.

aarmey:
I'd suggest adjusting Fig. 9, merging this, then doing the other corrections on a new branch.

Farnazmdi:
Thank you so much, I was really stuck. Will do right away.

Farnazmdi:
@aarmey I realized that although I am giving it the shape parameters as constant in the Emission function at the time of their import (in the `Lineage_collections.py`), the parameters are still being estimated. I know that Shakthi told me those emission functions that are used when we import, are dummy emissions, but I don't know what that actually means.

lineage/data/Lineage_collections.py
```diff
@@ -6,7 +6,7 @@
 
 #----------------------- Control conditions from both old and new versions -----------------------#
 desired_num_states = 2
-E = [StateDistribution() for i in range(desired_num_states)]
+E = [StateDistribution(shape1=22, shape2=10) for i in range(desired_num_states)]
```
aarmey, line 9: You don't want to set this on import, because you need to estimate them first.

lineage/figures/figure9.py
```diff
@@ -9,7 +9,18 @@
...
+PIs = []
+# to find the T and pi matrices to be used as the constant and reduce the number of estimations.
+for i in desired_num_states:
+    tHMM_solver = tHMM(X=Gemcitabine_Control, num_states=i)
+    tHMM_solver.fit()
+    # choose the estimated shape parameters for 1-state model to be kept constant
+    if i == 1:
+        constant_shape = [int(tHMM_solver.estimate.E[0].params[2]), int(tHMM_solver.estimate.E[0].params[4])]
+    Ts.append(tHMM_solver.estimate.T)
+    PIs.append(tHMM_solver.estimate.pi)
```
aarmey, line 25: tHMM uses whatever emissions are set within the data. So, here, you want to set them by `Gemcitabine_Control[i].E[j].G1.const_shape = const_G1`. You'll have to loop over every condition, lineage, state, and phase. It's weird to me that this is how the emissions are set, but it is what it is...

aarmey:
Also, see my second bullet point above. When there is censoring, the const parameter is being ignored. This has to be changed in the estimator, but I think you should do that on a new branch.

aarmey:
I'll add the const handling on a separate branch in a bit.

lineage/figures/figure9.py
Farnazmdi: I'm on it.

lineage/figures/figure9.py
```diff
@@ -9,24 +9,44 @@
...
+    # choose the estimated shape parameters for 1-state model to be kept constant
+    if i == 1:
+        constant_shape = [int(tHMM_solver.estimate.E[0].params[2]), int(tHMM_solver.estimate.E[0].params[4])]
+    for _ in range(len(data)):
+        Ts.append(tHMM_solver.estimate.T)
+        PIs.append(tHMM_solver.estimate.pi)
+
+for population in data:
+    for st in desired_num_states:
+        thmm_sol = tHMM(X=population, num_states=st)
```
aarmey, line 31: You are setting to a new tHMM object you then never use... I'll fix this right now.

aarmey:
It looks like the hanging came from raising the number of states to 7 from 4 max. The model scales *really* badly with increasing numbers of states, because essentially all the time is spent within the estimator. If we really have to we could possibly speed that up, but for now I've reduced this again so we can get this merged.

Farnazmdi:
@aarmey Are we sure that it is actually using the constant values that are given to it? I mean the transition and `pi` matrix. I don't seem them being passed to the `run_Analyze_over`.

aarmey:
I removed those. I'm not sure we want those fixed, and with the various recent improvements it looks like we are reliably seeing multiple states without those constraints.

Farnazmdi:
I see. Ok, thank you. I asked because I thought they were helping with the speed.

aarmey:
It worked! Merging.

PR #571, opened 2020-09-06
Plottree

Farnazmdi:
Adds 3 lineages of each drug below each graph.


Farnazmdi:
It  gets stuck somewhere in `Analyze()`. It seems it doesn't get passed line 52 @aarmey : `pred_states_by_lineage = tHMMobj.predict()`.
I wrote this based on everything in the master which builds. Do you have any clue what this might be from? I even reduced the number if lineages and states in figure 9, but still it gets stuck.

lineage/figures/figure9.py
```diff
@@ -29,18 +40,20 @@ def makeFigure():
     AICs = np.array([oo[0].get_AIC(oo[2], 4)[0] for oo in output])
     AICs = np.reshape(AICs, (desired_num_states.size, len(data)))
     AICs -= np.min(AICs, axis=0)
+    LAPlins = np.array([oo[0].X[0] for oo in output])[np.array([6, 7, 17])]
+    GEMlins = np.array([oo[0].X[0] for oo in output])[np.ayyay([-2, -3, -6])]
```
aarmey, line 44: array is mispelled.

aarmey (changes requested):
Where does that predict step happen within the code changed here?

aarmey:
I guess I'd suggest limiting the changes you're making relative to master. The only change you aim to make on this branch is adding the plots, but you've changed a bunch of other code. Reducing the amount of code changed will make it easier to find where the problem lies.

Farnazmdi:
The changes are somehow relative to figure 9, except for the changes in `Analyze()`.
In this PR:
1. up to 4 states are being plotted with different colors (in `plotLineage` file).
2. I had to add `height_ratios` to the `gridspec.GridSpec` in the `figurecommon` so I could somehow make the lineage trees that are plotted compact, without compacting the AIC figures themselves. (I wanted to show more lineages in the space that was given to one subplot)
3. added 3 tree plots for lapatinib and gemcitabine in `figure9.py`.

aarmey:
I think you're better off making a new figure rather than trying to mess with this file. Figure 9 does some extremely complicated things to fit all numbers of states, with parameters fixed, etc. If you simply plot lineages in a new file, you can fit one number of states, for one type of data at a time. This will make it easier to try things and quickly see whether it works.

PR #573, opened 2020-09-08
figure 10 plot lineages

Farnazmdi:
In this PR:
Figure 10 is added for plotting the lineage trees with their states shown in different colors.
There was a problem with the `plotLineage()` function that could not plot *some* those lineages with censoring. The length of cell's lifetime used to be handled by `cell.time.endT` and `cell.time.startT`. As I checked, it turned out that for some cells the `endT` was `NaN` and this was passed to the plotting function resulting in error.
I believe there is a problem with the way that `endT` is being calculated, but I wasn't sure where I should add a  few lines to make sure they won't become nan. I changed the code in the plotting function and set a couple of checks. Now it works.
I can spend some time in another branch and figure out where to fix the `endT` calculation.
For now, figure 10 shows 4 lineages of each drug.


aarmey (changes requested):
They all seem to be black?

Farnazmdi:
Yes sorry, just realized. I’m working on it in my local.

Farnazmdi:
image
@aarmey it builds now.

lineage/figures/figure10.py
```diff
@@ -0,0 +1,100 @@
...
+
+def figure_maker(ax, lapatinib, gemcitabine):
+    """
+    Makes figure 10.
+    """
+
+    ax[0].set_title("Lapatinib")
+    ax[1].set_title("Gemcitabine")
+
+    i = 0
```
aarmey, line 53: This looks a lot like a loop to me...

aarmey (approved):
Great. One change and then we can merge this.

I'm surprised the states don't look all that different. We can look at the fit emissions parameters to see what distinguishes them... maybe it's something cell cycle-specific.

PR #575, opened 2020-09-09
increased the number of states in AIC figure

Farnazmdi:
It took 700 seconds. I assume it is not that long.


Farnazmdi:
@aarmey Is there any package that I can look into to make this faster? However, relative to some other figures it doesn't seem that much longer.

aarmey:
I don't think so. We might be able to eek out a bit more performance by tuning the estimator, but I don't think it's worth out time. It's time to work on the paper.

aarmey (approved):

Farnazmdi:
Of course, I start to work on the results as soon as I finish the other PR. Thanks.

PR #576, opened 2020-09-09
Fixed Fig. 6

aarmey:
`all_cells` was accumulating all the lineages between each run, which broke the k-means results. This also adjusted the transition matrix and tolerances to better show the tHMM benefit.


Farnazmdi (approved):
The figure looks really good! Thank you very much!

Farnazmdi:
> `all_cells` was accumulating all the lineages between each run, which broke the k-means results. This also adjusted the transition matrix and tolerances to better show the tHMM benefit.

I thought it should accumulate them... I see why not now. Every data point has to be executed for itself. I knew I was doing differently from the other figures, should have known it's the wrong way. Thank you.

PR #587, opened 2020-09-11
discussion

Farnazmdi:
I made changes in the discussion.
Tried to increase the cell number in figure 7 and wanted to see if it reaches higher accuracy when in the 50-50% state proportion with emissions used in endpoints of figure 6. It didn't really improve. It may improve in other builds though, I will discard the changes in figure 7 if you say so. @aarmey


aarmey (changes requested):
The point of using `E2` is to have a common set of parameters we can describe when reporting the benchmarks. If you change those values it should be done in figureCommon, although I thought we set those to be reflective of the actual data.

Farnazmdi:
I don't think they were based on data.
It is a little hard to decide. Here are parameters of fitting both control conditions
image

Farnazmdi:
I can edit the `E2` to be:
state 0: [0.98, 0.98, 14, 2, 20, 2]
state 1: [0.99, 0.99, 14, 3, 40, 1]
Does it sound reasonable?

aarmey:
Does that resolve your issue?

Farnazmdi:
I was looking at the figures with this parameter set. Some of them look better.

aarmey:
Why not take the actual parameters from the two states fit with the data?

Farnazmdi:
They are different for gemcitabine and lapatinib. which ones should I take?

aarmey:
Maybe the fit from when you combine the controls for each drug?

aarmey:
Well actually... never mind. I think it's fine to use either one.

Farnazmdi:
image

figure 7 is worse than before.

aarmey:
What conditions were those for parameters from? They aren’t too different between states, which should make them hard to distinguish. I’d expect to get more distinct states with higher concentration of drug.

Farnazmdi:
> What conditions were those for parameters from? They aren’t too different between states, which should make them hard to distinguish. I’d expect to get more distinct states with higher concentration of drug.

Sorry, they were for control conditions. I tried a higher concentration.

Farnazmdi:
I tried them almost all, these are bad parameters. Builds keep failing. I'll bring those that used to be back.

PR #593, opened 2020-09-14
Fixed supp. figures

jclagarde:
Adjusted supp. figures to be consistent in size, font size, and cartoon size


Farnazmdi (approved):
Perfect!

PR #594, opened 2020-09-14
fixing the xaxis in pi error plots

Farnazmdi:
In this PR:
Fixed the x-axis of the `pi` estimation in the supplementary figures to be "Number of Lineages".
Made the changes to the `figueMaker` in the `figureCommon.py` so removed from the `figure5.py`. This way is more organized. 
The change made to the `commonAnalyze` was that I return the whole dictionary instead of some of its values. 
This change necessitated some edits to other figure files that used the `commonAnalyze` function.

In some files, I removed the `accuracy()` function as it seemed not very useful to have.


aarmey (approved):
Excellent!

PR #596, opened 2020-09-15
Emission

Farnazmdi:
plotted the G1 and G2 phase lengths for all concentrations in figure 11.

image

And here is for each concentration separately.
image

I can make them violin plots if that has more benefits.
2. Also edited the range of states in figure 10, it was wrong before. And replaced the colors with ones that are more distinguished. Now if you take a look at figure 10, it makes a little bit more sense.
These two are actually from the same fit, as I exported the concentration-wise results from figure 11 to figure 12.


Farnazmdi:
Here is for each concentration separately.
image

Farnazmdi:
I can make a table to show the distribution parameters, as well.

aarmey:
Try a stripplot: https://seaborn.pydata.org/generated/seaborn.stripplot.html

Farnazmdi:
I think it totally makes sense. In lapatinib, only G1 is separating the states and G2 is almost the same across the states, and in Gemcitabine both phases are improtant.
4E28CBB3-672F-4079-9D19-DF9A52CCCDB4
We don’t even need figure12 (separated concentrations)
I will make that table of parameters first thing in the morning.

aarmey:
You could try adding in multiple concentrations this way. Making the points smaller would also help to see all the points.

Screen Shot 2020-09-16 at 7 57 25 AM

Farnazmdi:
image
This looks much better!

aarmey:
Looks good, but I think all four concentrations are the same data?

lineage/figures/figure11.py
```diff
@@ -0,0 +1,80 @@
...
+gmc12_g1 = [] # for fig. 12
+gmc12_g2 = [] # for fig. 12
+
+# Run fitting
+output = run_Analyze_over(data, np.repeat([4, 3], 4))
+gemc_tHMMobj_list = [oo[0] for oo in output[0: 4]]
+gemc_states_list = [oo[1] for oo in output[0: 4]]
+lapt_tHMMobj_list = [oo[0] for oo in output[4: 8]]
+lapt_states_list = [oo[1] for oo in output[4: 8]]
+
```
Farnazmdi, line 29: There must be something wrong in these lines. I stared at them for more than half an hour but don't see anything wrong with it. All four populations for gemcitabine and all 4 populations in lapatinib are the same (as you said yersterday)

Farnazmdi: `output`'s length is 8 (which is correct), and each `output[i]` has the length of 3 (thmm_obj, states, likelihood).

aarmey: Have you printed the states lists and checked whether they are different? If they are the problem is after this point. If not it's before.

lineage/figures/figure11.py
```diff
@@ -0,0 +1,80 @@
...
+for idx, lapt_tHMMobj in enumerate(lapt_tHMMobj_list):
+    for lin_indx, lin in enumerate(lapt_tHMMobj.X):
+        for cell_indx, cell in enumerate(lin.output_lineage):
+            cell.state = lapt_states_list[idx][lin_indx][cell_indx]
+
+            if np.isfinite(cell.obs[2]): # g1 length is not nan
+                lpt_g1.append([cell.state, cell.obs[2]])
+            if np.isfinite(cell.obs[3]): # g2 length is not nan
+                lpt_g2.append([cell.state, cell.obs[3]])
```
aarmey, line 38: So, here I see you store the state paired with the observation.

lineage/figures/figure12.py
```diff
@@ -0,0 +1,49 @@
...
+    ax, f = getSetup((13.2, 6.66), (2, 4))
+    subplotLabel(ax)
+
+    concs = ["cntrl", "Lapt 25uM", "Lapt 50uM", "Lapt 250uM", "cntrl", "Gem 5uM", "Gem 10uM", "Gem 30uM"]
+
+    LAP = pd.DataFrame(columns=["state", "phase"])
+    GEM = pd.DataFrame(columns=["state", "phase"])
+    for i in range(4):
+        # lapatinib
+        LAP["state "+str(concs[i])] = [a[0] for a in lpt12_g1[i]] + [a[0] for a in lpt12_g2[i]]
```
aarmey, line 26: Where does this get filtered for state? Indexing operates on the position in a list, not on the value.

aarmey: Pretty sure the problem is here.

lineage/figures/figure11.py
Farnazmdi: image
this is the first 10 lists [state, G1_length] after appending them. It looks like they are the same from here on.

lineage/figures/figure12.py
aarmey: Wait... I'm not definitively sure this is wrong, but why don't you just make the Pandas table in the first place, rather than storing all these lists and then reformatting them?

lineage/figures/figure11.py
```diff
@@ -0,0 +1,80 @@
...
+
+def makeFigure():
+    """
+    Makes figure 11.
+    """
+    ax, f = getSetup((7, 3), (1, 2))
+    subplotLabel(ax)
+
+    Lap_data = pd.DataFrame(columns=["state", "phase lengths [hrs]", "phase"])
+    Lap_data["state"] = [a[0] for a in lpt_g1] + [a[0] for a in lpt_g2]
```
aarmey, line 64: You're doing the same repackaging here as below. In general, if you are repackaging data, and then repackaging it again, something is wrong. All these list loops are bad form.

lineage/figures/figure12.py
Farnazmdi: Hmm.. yes I didn't realize I am redoing it. I imported the values from the `fitting` in figure11.py so figure 11 and 12 to be from the same fit (not completely sure it is actually happening, though). I will create the dataframe in figure11 and import that.

lineage/figures/figure11.py
Farnazmdi: You're right, the reason I am creating these dataframes it because of `stripplot`. I know it is possible to pass the data in the form of lists, but didn't find a sensible example when a list is passed to `stripplot`.

lineage/figures/figure11.py
```diff
@@ -0,0 +1,80 @@
...
+output = run_Analyze_over(data, np.repeat([4, 3], 4))
+gemc_tHMMobj_list = [oo[0] for oo in output[0: 4]]
+gemc_states_list = [oo[1] for oo in output[0: 4]]
+lapt_tHMMobj_list = [oo[0] for oo in output[4: 8]]
+lapt_states_list = [oo[1] for oo in output[4: 8]]
+
+for idx, lapt_tHMMobj in enumerate(lapt_tHMMobj_list):
+    for lin_indx, lin in enumerate(lapt_tHMMobj.X):
+        for cell_indx, cell in enumerate(lin.output_lineage):
+            cell.state = lapt_states_list[idx][lin_indx][cell_indx]
```
aarmey, line 33: Why would the cell state be different than what's in the list?

lineage/figures/figure11.py
Farnazmdi: I think the screenshot below, shows the problem is before this point, am I right?

lineage/figures/figure11.py
Farnazmdi: cell states were not applied yet, I mean after fitting when printed cell states, they were `None`, so I had to assign them, but I think I don't need to necessarily assign them to `cell.state`, I can just use them...

aarmey: Yep.

lineage/figures/figure11.py
aarmey: Well, before the point you printed. There are multiple steps between there and here.

Farnazmdi:
Should be fixed now. I used lists instead of dataframes (they were easier than I thought).
I removed the plot that was showing the concentrations altogether and only kept separate concentrations. Let me know if you think it is helpful to have that-- although it doesn't seem to add any new information when we have separated ones.

lineage/figures/figure11.py
```diff
@@ -0,0 +1,63 @@
...
+    """ Makes figure 11. """
+
+    ax, f = getSetup((13.2, 6.66), (2, 4))
+    subplotLabel(ax)
+
+    # lapatinib
+    for idx, lapt_tHMMobj in enumerate(lapt_tHMMobj_list): # for each concentration data
+        lpt_g1 = []
+        lpt_g2 = []
+        for lin_indx, lin in enumerate(lapt_tHMMobj.X): # for each lineage list
```
aarmey, line 30: Can you make this duplicated code for lapatinib and gemcitabine a single function called twice?

aarmey (changes requested):
One edit below, and please make all the y-axis extend from 0 to 160.

This looks great. This plus a table of the fitted parameters (including T and pi) should be very informative.

aarmey:
Oh, also, I'm almost certain the drug concentrations are nM, not μM.

lineage/figures/figure11.py
Farnazmdi: Sure. About the parameters, these are dummy values. I wanted to ask you that should I just run the fitting once and print the parameters or should I write a function so in each build it overwrites the parameters?

aarmey: How are these dummy parameters? `run_Analyze_over()` in line 14 runs the fitting.

Farnazmdi: no no, I mean in the table in the methods.

aarmey: You don't need to put the values in the manuscript yet. Just collect them in a separate file so we can look through them. Depending on what they look like we can figure out how to present them.

Farnazmdi:
I think it is ready. I'll merge if you approve. I am preparing a file including the parameters and will send it to you soon. @aarmey

aarmey (approved):
Check that this builds but this looks good.

PR #600, opened 2020-09-18
Text edits

aarmey:
Still haven't gone through the first two paragraphs.


manuscript/01.introduction.md
```diff
@@ -6,15 +6,18 @@ One of the primary treatments of cancer consists of chemotherapy, mainly targete
...
+Measurements accompanied by lineage relationships are uniquely valuable for studying inherited phenotypes within families of people or populations of cells. This value is well-recognized in linkage studies which use pairs of relatives to identify the genetic determinants of disease [@pmid:19136655]. Notably, linkage studies can identify genetic determinants with greater power than even much larger correlation-based studies because relatives essentially serve as internal controls (CITE REVIEW). Linkage studies also start with the phenotype of individuals, rather than grouping based on molecular differences as is most common in cells. Lineage-resolved data has demonstrated unique value in cells for uncovering heterogeneity due to transient differences [@doi:10.1101/373258; @DOI:10.1073/pnas.1715639115].
```
Farnazmdi, line 14: Are either of these two suitable for the reference here?
[Kuchen et al](https://www.biorxiv.org/content/10.1101/373258v1.abstract)
[Chao et al](https://www.embopress.org/doi/full/10.15252/msb.20188604)

Farnazmdi (approved):

manuscript/01.introduction.md
aarmey: I do reference Kuchen et al. I'm not sure that Chao really demonstrates the value of having lineage relationships, but we should definitely reference it when we bring up the gamma-distributed times.

PR #603, opened 2020-09-20
manuscript update

jclagarde:
Added the emissions section in methods. I also went through the manuscript and tried to fix any typos that I found


Farnazmdi:
Great! Thanks, I will read through it as soon as I get a chance.

Farnazmdi (approved):

PR #604, opened 2020-09-20
another way of fit

Farnazmdi:
I'm progressing on the way you suggested to modify the fit and M-step of the fit, but here is my suggestion. 
I appended all the concentrations of lineages of a drug as a new population and fitted the model for 3 and 4 states (lapatinib and gemcitabine, respectively). This way I found the common transition and initial prob. matrices and used them as constant (`list_of_fT` and `list_of_fpi`) to find the emissions in populations of different concentrations.

The results make sense to me, but if you prefer that the BW part of the model to be modified, I will discard this and continue working on my other branch. @aarmey

### DIFF
--- a/lineage/Analyze.py
+++ b/lineage/Analyze.py
@@ -79,7 +79,7 @@ def run_Analyze_over(list_of_populations, num_states, parallel=True, **kwargs):
     list_of_fE = kwargs.get("list_of_fE", [None] * len(list_of_populations))
 
     if isinstance(num_states, (np.ndarray, list)):
-        assert len(num_states) == len(list_of_populations), f"len list population = {len(list_of_populations)}"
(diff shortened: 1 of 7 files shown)


aarmey:
The issue is that the emissions have to be concentration dependent. In this setup I’d bet that high concentrations just all end up as states that aren’t found at all in the control.

Farnazmdi:
The emissions are concentration-dependent. I just calculated the T and pi and used it to fit those separate populations. 
Or maybe I don't understand your point. Do you mean this way of calculating T and pi interferes with the way the emissions are calculated afterward?

aarmey:
But estimating T and pi *requires* estimating the emissions correctly for the same cells.

Farnazmdi:
Oh... I see. Ok, thank you. I will resume working on the other branch.

PR #605, opened 2020-09-20
figure1 updated

Farnazmdi:
Figure 1 updated based on our discussion, and the text changed accordingly.
image
I just noticed in the left part there are 10 cells. I will fix that right away.


aarmey (approved):
Excellent!

PR #608, opened 2020-09-21
Fit

Farnazmdi:
A new way of fitting.
Made `fit_list()` function in tHMM file that runs through a given list of data for concentrations.
Made `do_M_step_list()`, `do_M_pi_step_list()`, and `do_M_T_step_list()` functions in the BaumWelch file.
Made a new `Analyze_list()` function to be used instead of `run_Analyze_over` and made the corresponding changes to the figure10.py and figure11.py.

I ran the new `fit_list()` function for data for 3 concentrations in a list, and it worked for 2 states. I think it might be finished but it is very slow.


Farnazmdi:
@aarmey I built figure10 for 2 states and it was successful. It is just very slow, although I included the parallel execution from `Analyze()`. Do you mind taking a look at it to see if it could be faster in some way? @aarmey

lineage/tHMM.py
```diff
@@ -193,3 +193,52 @@ def log_E_score(EL_array, state_tree_sequence):
...
+    # sum over all populations in the list
+    total_new_LL = np.sum(new_LL)
+
+    # first stopping condition check
+    for _ in range(max_iter):
+        old_LL = total_new_LL
+
+        do_M_step_list(tHMMobj_list, MSD_list, betas_list, gammas_list)
+        for tHMM in tHMMobj_list:
+            MSD, NF, betas, gammas = do_E_step(tHMM)
```
aarmey, line 236: Doesn't this need to overwrite `MSD_list`, etc?

lineage/tHMM.py
```diff
@@ -193,3 +193,52 @@ def log_E_score(EL_array, state_tree_sequence):
...
         log_E_score_holder += row[state_tree_sequence[idx]]
     return log_E_score_holder
+
+
+def fit_list(tHMMobj_list, tolerance=1e-9, max_iter=1000):
+    """Runs the tHMM function through Baum Welch fitting for a list containing a set of data for different concentrations"""
+
+    # Step 0: initialize with KMeans and do an M step
+    # when there are no fixed emissions, we need to randomize the start
+    init_gammas = [[sp.multinomial.rvs(n=1, p=[1. / tHMMobj_list[0].num_states] * tHMMobj_list[0].num_states, size=len(lineage))
```
aarmey, line 203: I don't think this line does anything?

aarmey (changes requested):
Great! Found one error, but that wouldn't explain slow fitting. You can run `python3 -m cProfile -o profile genFigures.py 10` and then `gprof2dot -f pstats --node-thres=5.0 profile | dot -Tsvg -o profile.svg` to profile figure 10. Maybe that might highlight a slow spot?

lineage/tHMM.py
Farnazmdi: Corrected.

lineage/tHMM.py
Farnazmdi: Removed.

lineage/tHMM.py
```diff
@@ -232,9 +229,13 @@ def fit_list(tHMMobj_list, tolerance=1e-9, max_iter=1000):
         old_LL = total_new_LL
 
         do_M_step_list(tHMMobj_list, MSD_list, betas_list, gammas_list)
-        for tHMM in tHMMobj_list:
+        for idx, tHMM in enumerate(tHMMobj_list):
             MSD, NF, betas, gammas = do_E_step(tHMM)
             new_LL.append(np.sum(calculate_log_likelihood(NF)))
```
aarmey, line 234: Do you ever empty the `new_LL` list for the next fitting iteration?

Farnazmdi: Oh.. I guess not. Will do it right away. Thanks

Farnazmdi:
> gprof2dot -f pstats --node-thres=5.0 profile | dot -Tsvg -o profile.svg

It doesn't show anything. I guess it is not that bad. It is running through all the concentrations every time instead of one of them.
image

aarmey (approved):

PR #609, opened 2020-09-22
Emissions

Farnazmdi:
I made a few edits to your text for emissions and created a table for the parameters.
 We probably will change the parameter sets and then we will have a known and reasonable source, let's keep it for now (I wrote 5 nM lapatinib).


jclagarde:
just saw this! looks good! thank you for the edits! let me know if there is anything else I can do :)

PR #611, opened 2020-09-22
figure11 some addition

Farnazmdi:
In this PR, 4 subplots were added to figure 11 to show the average phase lengths for different states versus the concentrations, with the average phase lengths being `shape \times scale`. I thought this might be a good visualization.

@aarmey I remember you mentioned yesterday that we can ultimately have cell numbers versus concentrations, something like what we have in the ODE model. I may have heard wrong and you said something else, but if I heard correctly do you mean that we have a specific time point and calculate cell number in different concentrations for that specific time point and plot versus the concentration?


aarmey:
No, I was only talking about phase lifetime (scale * shape) like it sounds you have here.

Farnazmdi:
image
Please let me know if anything needs to change.

aarmey:
Can you add a plot/plots for the Bernoulli parameters?

Farnazmdi:
Sure.

aarmey (approved):
This looks good to merge.

aarmey:
There's some tweaking you can do with the axes, etc, but the basic output is working well.

Farnazmdi:
I will ask about them at the next meeting. Thanks.

PR #612, opened 2020-09-23
Consolidate tHMM methods

aarmey:
This uses the list methods in both cases by just putting the tHMM object into a list with one item. I also switched the XLSX reader to avoid a deprecation. In the fitting process, we now exit when the fit stops improving, not when it stops changing. This was causing the fitting to run for a long time without getting any better. I think it only showed up now because we have many more cells in one fitting run.


aarmey:
Build back down to ~20 min. :-)

PR #626, opened 2020-09-27
fig12

Farnazmdi:
This PR:
time-until-death distribution for all concentrations in figure 12.
and the transition matrix in figure 13.

The distribution of these times is not clear to me, not sure if it helps.

### DIFF
--- a/lineage/figures/figure12.py
+++ b/lineage/figures/figure12.py
@@ -3,91 +3,50 @@ import numpy as np
 import seaborn as sns
 import networkx as nx
 
-from .figure11 import gemc_tHMMobj_list, lapt_tHMMobj_list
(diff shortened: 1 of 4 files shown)


Farnazmdi:
Networkx is a little bit out of control. I kept changing the dimensions and sizes, this is the best I could get so far. I tried to control the length of the arrows, but they wouldn't change! If I choose a smaller grid size, the circles that show the states will be cut from the sides.

aarmey:
Excellent! I agree, the numbers of cells are pretty small to identify the distribution shape. Maybe a simpler strategy would be to keep the emissions structured the same as we have them, with a Gamma distribution and Bernoulli parameter, but simply remove the cell death times from the Gamma estimation and probability calculation? This would be equivalent to not caring about the amount of time a cell takes to die.

aarmey (approved):

PR #628, opened 2020-09-28
Optionally set pi by T

aarmey:
This adds an optional argument to set pi using T. Note that this does not take into account the influence of emissions on the distributions of cell states.


Farnazmdi:
Don't numbers like this 0.006 here mess with the assumption of irreducibility?
image

aarmey:
Don't think I understand?

aarmey:
What do you mean by irreducibility?

Farnazmdi:
I mean the irreducibility of the Markov chain is a requirement to have a unique stationary distribution, which I think means we must be able to go from any state to any other state with a positive probability. What I thought might be concerning is that sometimes I have seen that the AIC figure picks more number of states as optimum and sometimes few of the elements in the T matrix were zero (not sure if we are going to set the number of states to automatically choose the optimum one from the AIC results...)
Sorry... Nevermind. Even if one of the indices of T is zero, we still can reach that state through another state. Sorry now that I thought about it, there is no concern.

PR #632, opened 2020-09-28
Edit dist estimation

Farnazmdi:
Removed dead cells from being accounted for in gamma parameter estimation. Done is a new mask right before when we were removing those that were `nan`.


lineage/CellVar.py
```diff
@@ -58,6 +58,12 @@ def isLeafBecauseTerminal(self):
         # otherwise, it has no left and right daughters
         return True
 
+    def hasDied(self):
+        """ Returns true if the cell has died. 
+        Mostly used for removing them while estimating Gamma distributions. """
+        if self.obs[0] == 0 or self.obs[1] == 0:
```
aarmey, line 64: This depends on the emissions distribution, and so should not be defined here.

lineage/states/StateDistributionGamma.py
```diff
@@ -72,16 +72,22 @@ def estimator(self, x, gammas):
         γ_obs = x[:, 1]
         gamma_obs_censor = x[:, 2]
 
+        # removing those who died from the gamma list
+        gamma_death_mask = bern_obs == 1
```
aarmey, line 76: Simplified this. There's no need for a loop.

lineage/states/StateDistributionGamma.py
```diff
@@ -72,16 +72,22 @@ def estimator(self, x, gammas):
         γ_obs = x[:, 1]
         gamma_obs_censor = x[:, 2]
 
+        # removing those who died from the gamma list
```
aarmey, line 75: pdf() will correspondingly need to change.

aarmey (changes requested):

lineage/states/StateDistributionGamma.py
```diff
@@ -48,16 +48,22 @@ def pdf(self, x):  # user has to define how to calculate the likelihood
...
         # Update censored Gamma
-        ll[x[:, 2] == 0] += sp.gamma.logsf(x[x[:, 2] == 0, 1], a=self.params[1], scale=self.params[2])
+        ll_g[x[:, 2] == 0] += sp.gamma.logsf(x[x[:, 2] == 0, 1], a=self.params[1], scale=self.params[2])
+
+        # remove dead cells
+        ll_g[x[:, 0] == 0] = 0
+
+        ll= ll_b + ll_g
 
         return np.exp(ll)
```
Farnazmdi, line 68: Is this a correct change for the pdf function? @aarmey

Farnazmdi: For the dead cells we only have the bernoulli likelihood, so I set the likelihood calculated from the gamma to zero.

aarmey (changes requested):
I cleaned things up some. Something is broken but I'm not seeing it... when I printed out the LL during each iteration, it was getting worse over the course of the fit. Could you take a look?

Farnazmdi:
Thank you. Sure, will do.

Farnazmdi:
I checked the branch master and the likelihood for a fit was getting worse, too. I don't think it is in this branch, I checked many things. I need to look closer.

Farnazmdi:
The test is failing wit ha small difference 
image
I think since after masking the dead cells (censoring time) we lose more than half of the cells, it is safe to increase the tolerance by a little amount. Clearly, the average of the time lengths (shape x scale) has improved (in the defined emission it is 7 x 4.5).
@aarmey  Should I relax the assertion bound a little bit?

aarmey:
In your logical_and example, I'm not sure that's showing it doesn't work as expected. `a` in your example is a list, while it's a NumPy array in the estimator (I think).

Farnazmdi:
Oh.. yes, you are right..
I will revert that

lineage/states/StateDistributionGamma.py
```diff
@@ -73,8 +76,10 @@ def estimator(self, x, gammas):
...
 
         b_mask = np.isfinite(bern_obs)
-        g_mask = np.isfinite(γ_obs)
-
+        # Both unoberved and dead cells should be removed from gamma
+        g_mask = np.logical_and(np.isfinite(γ_obs), bern_obs == 1)
+        # at least have 10 cells after masking
+        assert np.count_nonzero(g_mask) >= 3
         # Handle an empty state
         if np.sum(gammas[b_mask]) == 0.0:
```
Farnazmdi, line 84: Is this assertion reasonable? I mean, do we expect the tests to always pass if there are at least 3 cells for estimation? I made it 10 first but 9 tests failed! @aarmey

aarmey: Yes. One of the failures seems to be from all the cells being dead and therefore eliminated. Set the emissions in test_BW so that there's essentially never cell death.

PR #633, opened 2020-09-29
transition matrix

Farnazmdi:
@aarmey it works on notebook, 
image

but when I add the `pygraphviz` to the requirements (and installed it by `pip install pygraphviz`) it errors:
Do you have an idea what might be wrong?

image


aarmey:
Don't bother getting it to build right now. A notebook for this specific plot is fine.

PR #646, opened 2020-10-05
Added Supp. Figure Captions

jclagarde:
I'm not quite sure where the text explanation that was previously in the caption section should go. I added it above the figure though.


Farnazmdi:
@JC Lagarde I have seen many papers without explaining their supplementary figures, and only providing a caption. Besides, we have mentioned these figures in the main results where it was relevant. @aarmey Do we want a paragraph explaining each supplementary figure?

aarmey:
@Farnazmdi you're right that the captions likely should do this. Optionally, we could include a supplementary methods section with a more detailed description.

Farnazmdi (approved):

PR #654, opened 2020-10-09
Delete figure3.svg

Farnazmdi:
Some of the edits related to the slack thread.
Removed figure3, but kept the text in the results.

Moved figure 7 to supp. and commented the text. I wasn’t sure we want to keep the text corresponding to it or not. Also commented out figure S05 (censored lineage simple gamma obs. for relative presence of states)

Added Drs. Heiser and Gross, but I couldn’t find their github and Dr. Heiser’s orcid id. We could ask them on Monday.


manuscript/20.supp-figures.md
```diff
@@ -24,14 +24,14 @@
...
+<!-- ![**Performance on populations of censored lineages of increasing number.** (a) Visualization of the proportion of cells in the censored lineage transitioning from state 1 to state 2. (b) Bernoulli parameter deviates from the true value as percent of cells in state 1 increases for state 1 and approaches the true value when the percent of cells in state 1 is above 40% for states 1. (c)-(d) Gamma shape and scale parameters approach the true value then deviates when over 80% of cells are in state 1 for states 1 and 2. (e) State assignment accuracy decreases as the percent of cells in State 1 increases. (f) Errors in estimating the transition probability matrix, T, approaches 0 when 40-75% of cells are in state 1 and increases at other percentages for states 1 and 2.](./output/figureS05.svg){#fig:real_5} -->
```
aarmey, line 34: Is this meant to be commented out?

manuscript/metadata.yaml
```diff
@@ -21,6 +21,12 @@ authors:
     orcid: 0000-0002-0197-4670
     affiliations:
       - Department of Bioengineering, University of California, Los Angeles
+  -
+    name: Sean M. Gross
```
aarmey, line 25: We can just leave out their GitHub and/or ORCID ID.

manuscript/20.supp-figures.md
aarmey: Hmm... can you briefly summarize how each of the three figures are different? I don't want to remove it just because it's weird, but I agree we can reduce these down.

Farnazmdi: Figure 7: tree plot to show censored vs uncensored, and shows only state assignment accuracy for the two cases, simulated for phase-specific emissions.
Figure S04: Shows all the estimation (accuracy, parameter estimation, T and pi errors) for uncensored lineages with simple gamma emissions
Figure S05: Same as S04 only for censored lineages.

Farnazmdi: They all have overlap...

aarmey: Ok, let's get rid of 7 but keep S04 and S05.

PR #661, opened 2020-10-10
edited plots fig 11

Farnazmdi:
Only changed 2 lines to keep cells that were not censored. Here is the error:
image

I didn't save these values in a dataframe, they are only lists. One of the reasons I found online for this error was that the name of two columns in a dataframe were the same, but not the case here.


aarmey:
Your code right now assumes that g1 and g2 are the same length. I *think* seaborn will just not plot a point if it's NaN, so you could append a NaN if the phase is censored?

Farnazmdi:
Thank you.

aarmey (approved):

PR #667, opened 2020-10-17
increased number of points for violin plot

jclagarde:
@aarmey 100 data points vs 300 data points for the violin plots 
Screen Shot 2020-10-16 at 7 27 07 PM
Screen Shot 2020-10-16 at 7 22 46 PM


aarmey:
We just want more data points in the violins. Not in the accuracy plot.

Farnazmdi (approved):

PR #670, opened 2020-10-19
Baseline for Wasserstein figure

jclagarde:
Run lineage using 1 state in order to get baseline value


lineage/figures/figure5.py
```diff
@@ -16,6 +16,11 @@
     commonAnalyze
 )
 from ..LineageTree import LineageTree
+from ..states.StateDistributionGamma import StateDistribution
+
+pi2 = [1]
+T2 = np.array([1])
+E3 = [StateDistribution(0.99, 8, 6)]
```
Farnazmdi, line 24: replace them with:
`pi = np.ones(len(E3)) / len(E3)
T = np.eye(len(E3)`
and define E above them.

Farnazmdi: make sure to `git pull`

jclagarde:
I think that trying to do 1 state runs into lots of issues with the code. I tried commenting this out as well as other sections that refer to two states 

Screen Shot 2020-10-18 at 9 00 58 PM

Farnazmdi:
I don't think it is lots of issues. You only need to add an if statement above line 136.
`if len(tHMMobj.X[0].E) == 2:`
 `    <<line 136...>>`
`else:`
`    pass`.

Farnazmdi:
You shouldn't touch the way the population is being created, and shouldn't touch the last line for now.

PR #671, opened 2020-10-19
FigureS04

jclagarde:


jclagarde:
@aarmey Some things I have tried with the transition matrix in figureS04

PR #672, opened 2020-10-19
Figs04 05

Farnazmdi:
1. The keyword for `kwargs` for passing `list_of_Ts` in `commonAnalyze` was not correct.
2. changed bernoulli for state 2 from 0.75 to 0.9.
3. T is now only changing from 0.6 to 0.99 and `pi` is being estimated. 
The difference with the previous figure is that we only see half of the previous figure, meaning we see that the proportion of cells in the two states changes from ~60% to ~100% and this results in a slight decrease of accuracy which is expected and I think it makes the point.

### DIFF
--- a/lineage/figures/figureCommon.py
+++ b/lineage/figures/figureCommon.py
@@ -24,7 +24,7 @@ pi = calculate_stationary(T)
 
 # bern, gamma_a, gamma_scale
 state0 = StateDistribution(0.99, 8, 6)
-state1 = StateDistribution(0.75, 8, 1)
+state1 = StateDistribution(0.9, 8, 1)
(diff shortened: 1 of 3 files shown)


Farnazmdi:
Figure S05:
image

Figure S04:
image

Farnazmdi:
The build fails after 14 seconds... I am not sure what's going on.

aarmey:
Should be fixed. Rerun.

aarmey (approved):
Check the results once this builds, but if it looks reasonable it should be ready to merge.

Farnazmdi:
Wow! remarkible!
image

PR #678, opened 2020-10-21
Excel

Farnazmdi:
Saving cells' state in a tree manner in an excel sheet, similar to the input data.
I think if `xlsxwriter` works, this should be ready, but it still doesn't recognize this package.


lineage/data/Lineage_collections.py
```diff
@@ -8,38 +8,42 @@
...
-def popout_single_lineages(lineages):
-    """ To remove lineages with cell numbers <= 5. """
-    trimed_lineages = []
-    for cells in lineages:
-        if len(cells) < 5:
-            pass
-        else:
-            trimed_lineages.append(cells)
-    assert len(trimed_lineages) > 0
+def popout_single_lineages(lineages):
```
aarmey, line 11: Do we use this?

lineage/data/Lineage_collections.py
```diff
@@ -8,38 +8,42 @@
...
+len_lp_cntr = [len(lap01), len(lap012), len(lap02), len(lap03)]
 
 # -- LAPATINIB 25 uMolars
 
 lapb63 = [LineageTree(list_of_cells, E) for list_of_cells in import_Heiser(path=r"lineage/data/heiser_data/new_version/AU00601_B6_1_V4.xlsx")]
-lapb64 = [LineageTree(list_of_cells, E) for list_of_cells in import_Heiser(path=r"lineage/data/heiser_data/new_version/AU00701_B6_1_V4.xlsx")]
-lapb65 = [LineageTree(list_of_cells, E) for list_of_cells in import_Heiser(path=r"lineage/data/heiser_data/new_version/AU00801_B6_1_V4.xlsx")]
-# replicates
 lapb632 = [LineageTree(list_of_cells, E) for list_of_cells in import_Heiser(path=r"lineage/data/heiser_data/new_version/AU00601_B6_2_V4.xlsx")]
+lapb64 = [LineageTree(list_of_cells, E) for list_of_cells in import_Heiser(path=r"lineage/data/heiser_data/new_version/AU00701_B6_1_V4.xlsx")]
```
aarmey, line 38: Why does this require changes to the file import?

lineage/export_states.py
```diff
@@ -0,0 +1,146 @@
+""" This file exports excel sheets of lineages of cells with their states assigned to them after fitting. """
```
aarmey, line 1: How much do the results change here from fitting run to fitting run?

lineage/data/Lineage_collections.py
Farnazmdi: It is used for figure 1. I can go through the files and find proper lineages, but for now the indexes we use to plot those exact lineages.

lineage/data/Lineage_collections.py
Farnazmdi: I made changes to the order of appending them so I won't lose track of the files that which is which. Also found one or two things needed modifications, I guess in one of them, I saw a repetition of the same file so I changed it, and in another case, found that I hadn't included a file or two, and added them.

lineage/export_states.py
Farnazmdi: Honestly, I don't know, but I think it is a lot. I set a random seed in figure 11, too, so it wouldn't change, but haven't set it in the file that runs the fitting and creates this export.

aarmey (approved):
Ok, this code all looks good.

I do not recommend setting the random seed. It would be better to understand the variation between fits. It's very likely that some of those fits are good and others are bad—setting the random seed potentially ensures you only see bad ones.

Farnazmdi:
Thank you. I will remove the lines where I set the random seed. 
I'll try to compare different runs and see what I can understand. If we keep getting different results from run to run, it will be difficult to write interpretations in the manuscript.

PR #681, opened 2020-10-25
figure edits and added result part for wass

Farnazmdi:
1. Some cleanup in figure1, and added titles for the lineages to show the concentration they were treated with.
2. Added a sentence in the result to show a baseline for Wasserstein.


aarmey (changes requested):
Sorry, I meant the conditions should go in the figure caption, not the figure itself. If placed as the title, I think it implies that we expect the difference is drug-induced.

aarmey (approved):

PR #685, opened 2020-10-27
transition matrix figure

Farnazmdi:
@aarmey I have the same problem as JC's PR. But I haven't cloned the repo inside itself.


aarmey:
It's because as soon as that build fails in that way, the build server doesn't know how to remove the submodule. I'll clean it out.

Farnazmdi:
Thank you very much.

aarmey (changes requested):
We need to add arrows indicating when a state stays itself.

Farnazmdi:
Actually, that's what I'm struggling with right now. It looks like `draw_networkx_edges` doesn't do that. I'll look more.

aarmey:
Darn. Looks like it is indeed not an option. In that case I think we're stuck using graphviz.

Farnazmdi:
:-( Should I save the `.svg` and overlay on the manuscript?

aarmey:
Yep

PR #687, opened 2020-10-27
added the transition

Farnazmdi:
image
image

@aarmey The dependency installs have the same error as I used to get before you installed `pygraphviz` on the server.


aarmey (changes requested):
Think I just fixed that. The numbers on the graph start at 1 (which I think is better), while the state numbers in the plots start at 0. Can you make these consistent?

Farnazmdi:
Absolutely, will do right away.

Farnazmdi:
@aarmey The installations for tests fails for the same reason. Could you please let me know how you fixed that for the build? 
image

aarmey:
I think I had just rerun build but not test.

lineage/figures/cartoons/gemcitabine.svg
```diff
@@ -0,0 +1,132 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
+ "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
aarmey, line 3: If these are programmatically generated, they shouldn't be included in the repo.

aarmey (changes requested):
Ok, I think I've finally fixed this.

lineage/figures/cartoons/gemcitabine.svg
Farnazmdi: Thanks so much.
Yes, they are generated. Is there a way to remove them after they are called in `genFigures`?

aarmey: Look for how you can delete a file using Python.

aarmey (approved):

PR #698, opened 2020-11-11
fit all concentrations at once

Farnazmdi:
Here is so far:
1. Wrote new functions in `stateCommon.py` for estimating the gamma parameters for all concentrations at once. 
2. Made a new class named `StateDistAll` in the `StateDistributionGamma.py` that accumulates the state distributions for all the concentrations. 
3. Made a new class named `AllObjs` in the `tHMM.py` to collect all the tHMM objects for different concentrations. 
4. In the  `M_E_step` in `BaumWelch.py` made an if statement that if the tHMMobj that is given is a list (which means it is for all concentrations at once) do a new method using those two newly made classes, otherwise does the old way of `M_E_step`.

Do you see a redundant step in these 4 that I have done @aarmey ? I'm actually a bit stuck and not sure if this is the right way. I will make a test to see if it works.


aarmey:
I don't think we need a new state distribution; I think this can be accomplished within the structure of the current one. The emissions classes are a means for *storing* information, while we want the *behavior* to be different. I think the only two things we need are (1) a method that collects a list of gammas and observations across concentrations, and (2) a function that will take these lists and then perform the gamma distribution estimation. After the latter runs, the gamma distribution results can just be copied back into the current emissions distributions.

Maybe try writing #1 first? The second is essentially using the same approach to optimization of the distribution likelihoods, except while setting a [linear constraint](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.LinearConstraint.html#scipy.optimize.LinearConstraint) on the parameters between concentrations.

Farnazmdi:
Will do it this way, thank you.

Farnazmdi:
Sorry, just to clarify, for (1) you mean a method in the tHMM class, right? Because we won't access the statedistribution class anywhere after importing the lineages.

aarmey:
It'd probably need to be outside of the tHMM class, but take a list of tHMM objects, since we have one tHMM per condition.

Farnazmdi:
If we keep everything the same except that we gather the list of tHMM objects and for each of them from control to concentration_1, ... to concentration_last we pass the scale_estimated from the previous concentration to the current one that is being fitted, we don't need the linear_contraint and we don't change anything much. The conditions run separately but we control the upper bound of the scale parameter. How does this sound? Although, with this, we have to stick with the way we estimate shape_parameter.

aarmey:
You have to fit them all at once because doing it step-wise makes the first condition special, which it's not.

Farnazmdi:
I know I added many lines, but I can't figure out which part is not working... It is not being included in the fitting although I made sure the proper functions are being executed when we have more than 1 tHMMobj in the list that is being passed to the analyze_list. I really appreciate if you could take a quick look and see if something seems off.

lineage/BaumWelch.py
```diff
@@ -145,6 +150,31 @@ def do_M_E_step(tHMMobj, gammas):
         tHMMobj.estimate.E[state_j].estimator(all_cells, all_gammas[:, state_j])
 
 
+def do_M_E_step_atonce(all_tHMMobj, all_gammas):
+    """ perform the M_E step when all the concentrations are given at once. """
+    all_cells = []
```
aarmey, line 155: You should have this all work for one state at a time. Collecting everything across states makes it more complicated.

lineage/BaumWelch.py
```diff
@@ -145,6 +150,31 @@ def do_M_E_step(tHMMobj, gammas):
...
+    final_gm = []
+    for j in range(all_tHMMobj[0].num_states):
+        tmp1 = []
+        for aray in gms:
+            tmp1.append(aray[:, j])
+        final_gm.append(tmp1)
+
+    # for each state, estimate the parameters and assign them to the tHMMobj
+    for state_j in range(all_tHMMobj[0].num_states):
+        output = atonce_estimator(all_cells, final_gm[state_j])
```
aarmey, line 172: I'd check that you're collecting all of the data together, and then assigning the results, correctly before worrying about the estimator. You can do this by simply skipping the actual estimator call and assigning the same parameter values as were there before.

lineage/states/StateDistributionGamma.py
```diff
@@ -190,3 +189,46 @@ def time_censor(cell, desired_experiment_time):
             # the daughters are no longer observed
             cell.left.observed = False
             cell.right.observed = False
+
+def atonce_estimator(x_list, gammas_list):
```
aarmey, line 193: You know what the input and output to this should look like. You should write this on a separate branch, and then write a test that checks it works.

lineage/states/StateDistributionGamma.py
```diff
@@ -190,3 +189,46 @@ def time_censor(cell, desired_experiment_time):
...
+    """ User-defined way of estimating the parameters given a list of the tuples of observations from a group of cells. """
+    # unzipping the list of tuples
+    x_data = []
+    for x in x_list:
+        x_data.append(np.array(x))
+
+    assert len(x_data) == len(x_list)
+    # getting the observations as individual lists
+    # {
+    bern_obs = [x[:, 0].astype('bool') for x in x_data]
```
aarmey, line 203: Bernoulli shouldn't be handled in here. We could add constraints on it, too, but not right now. It can be handled by calling the normal estimators. Then, the constrained gamma solution can overwrite the normal gamma estimator.

lineage/states/stateCommon.py
```diff
@@ -75,6 +84,42 @@ def gamma_estimator(gamma_obs, time_cen, gammas, constant_shape, x0):
 
     return xOut
 
+def gamma_estimator_atonce(gamma_obs, time_cen, gamas):
```
aarmey, line 87: You know what the inputs and outputs to this should look like. This should be written with a separate branch and test.

aarmey (changes requested):
Here are some thoughts in case it's not working. However, the real problem is manubot and something I need to fix. I'll make a separate PR.

aarmey:
Ok, the manuscript build problem should now be fixed.

lineage/BaumWelch.py
Farnazmdi: Do you mean that the functions should be written for one state or you mean I write a for loop over the state inside this function and do everything for each step at each iteration?

lineage/states/StateDistributionGamma.py
Farnazmdi: I moved this to another branch.

lineage/states/StateDistributionGamma.py
Farnazmdi: removed the bernoulli estimation from the estimator (that is in another branch)

lineage/states/stateCommon.py
Farnazmdi: Moved to another branch.

lineage/BaumWelch.py
Farnazmdi: Should I collect all the cells and all the gammas when I am running the functions on each state? Or should I only pass those cells and those gammas for each specific state? I thought I should pass all and the gammas are there for distinguishing the states... I'm not sure I'm right

lineage/BaumWelch.py
aarmey: I mean you should have a function that concerns itself with just one state. You can then call it N states numbers of times.

lineage/BaumWelch.py
aarmey: Functions should only receive the data they need for the action at hand.

lineage/BaumWelch.py
Farnazmdi: Thank you.

Farnazmdi:
Here is in this branch:
1. A unit function in the `BaumWelch.py` that takes in all the cells and gammas for one state, and is supposed to run the estimator and assign the parameters.
2. A function that calls the unit function for each state and it called `do_M_E_step_atonce`.
3. A slight adjustment in the `do_M_step` and in the `fit_list()` in the tHMM, so when we are passing more than one concentration it runs the `...at_once` functions automatically.
I have the estimators in another branch.

lineage/figures/cartoons/gemcitabine.svg
```diff
@@ -1,93 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
- "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
aarmey, line 3: Did you intend to delete this?

lineage/figures/cartoons/lapatinib.svg
```diff
@@ -1,93 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
- "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
aarmey, line 3: Did you intend to delete this?

lineage/tHMM.py
```diff
@@ -178,11 +177,16 @@ def fit_list(tHMMobj_list, tolerance=1e-9, max_iter=1000):
...
-                       for lineage in tHMM.X]
-
-        do_M_E_step(tHMM, init_gammas)
+                        for lineage in tHMM.X]
+        init_all_gammas.append(init_gammas)
+    if len(tHMMobj_list) > 1:
+        do_M_E_step_atonce(tHMMobj_list, init_all_gammas)
+    else:
+        for i, tHMM in enumerate(tHMMobj_list):
+            do_M_E_step(tHMM, init_all_gammas[i])
```
aarmey, line 189: Don't you need to do this for `do_M_step` on line 197, too?

aarmey: Honestly I don't remember why we have both `do_M_step` and `do_M_E_step`... I might be forgetting something critical but it'd be nice if they were combined or one just called the other after some setup.

lineage/BaumWelch.py
```diff
@@ -145,6 +148,33 @@ def do_M_E_step(tHMMobj, gammas):
...
+            tHMMobj.estimate.E[state_j].params[i] = 0.5
+
+def do_M_E_step_atonce(all_tHMMobj, all_gammas):
+    """ perform the M_E step when all the concentrations are given at once for all the states.
+    gms is a list of arrays with size = N x K.
+    After reshaping, we will have a list of lists for each state. 
+     """
+    gms = []
+    for gm in all_gammas:
+        gms.append(np.vstack(gm))
```
aarmey, line 169: Why `vstack` one array?

aarmey (changes requested):
A few things.

lineage/figures/cartoons/gemcitabine.svg
Farnazmdi: Yes, actually, these are generated every time we run the whole figures. I know I had written something in the makefile to delete them each time, but not sure why it is not doing that.

lineage/figures/cartoons/lapatinib.svg
Farnazmdi: same as above.

lineage/tHMM.py
Farnazmdi: Because the `do_M_step` takes the list of tHMM objects, I wrote this if statement inside it, but the `do_M_E_step` only applies on one tHMM object at a time, so I had to keep the if statement here. But you are right, there is no need to separate them. I will combine the two right away.

lineage/BaumWelch.py
Farnazmdi: In this, I followed the pattern that was being done for each population, only here I have the data of concentrations appended in a list, so I looped over those inside the list. I don't think it is one array; it should be the matrix that is keeping the N x K gammas for each lineage and this should be stacking all the lineages of the same condition.

lineage/tHMM.py
Farnazmdi: I removed the `do_M_E_step` but didn't removed the one for `atonce_...`.
Also, it seemed to me that the initialization in the `fit_list()` was not really doing anything and I removed it and built a few of the figures, they were not different at all.
Is it necessary to have random initialization before the first step of EM? I think because we initialize everything in the classes this may not be necessary...

aarmey: This all sounds reasonable. Is this ready for review then?

Farnazmdi: It turned out that by removing the initialization, apparently, it messed with calculating the `lowess` for figure S04. I searched the error (which was this: `FloatingPointError: invalid value encountered in true_divide`) and it doesn't seem to be easily fixable! So brought the `do_M_E_step` back so we could call only that in the initialization part in `fit_list`.

Farnazmdi:
Sorry, I had forgotten to re-request your review on this.

aarmey (commented):
This all looks good to me. Let's leave this here, and then we will add the other parts by making PR's onto this branch. Finally, this will get merged with master when all the parts are done.

Farnazmdi:
Sure, sounds great. The estimator PR is close to ready, there are few issues that I need to figure out.

lineage/states/stateCommon.py
```diff
@@ -105,3 +102,41 @@ def basic_censor(cell):
...
+
+def negative_LL_atonce(x, uncens_obs, uncens_gammas, cens_obs, cens_gammas):
+    """ uses the negative_LL_atonce and passes the vector of scales and the shared shape parameter. """
+    outt = 0.0
+    for i in range(4):
+        outt += negative_LL_sep(x[1 + i], x[0], uncens_obs[i], uncens_gammas[i], cens_obs[i], cens_gammas[i])
+    return outt
+
+
+def gamma_estimator_atonce(gamma_obs, time_cen, gamas, x0=None):
```
aarmey, line 115: @Farnazmdi I think what will help more than anything else is to pass the parameters from the last iteration in as a starting point (`x0`). Now, the first iteration won't satisfy the bounds, so you could write a check in here to override the starting point if it isn't valid. We do this with the other estimator, and it means that the first iteration takes the same amount of time, but subsequent ones are much faster.

Farnazmdi: Sorry, this had slipped out of my sight. I'm on it.

lineage/states/StateDistributionGamma.py
```diff
@@ -190,3 +189,35 @@ def time_censor(cell, desired_experiment_time):
...
+
+    # Both unoberved and dead cells should be removed from gamma
+    g_masks = [np.logical_and(np.isfinite(γ_o), berns) for γ_o, berns in zip(γ_obs, bern_obs)]
+    for g_mask in g_masks:
+        assert np.sum(g_mask) > 0, f"All the cells are eliminated from the Gamma estimator."
+
+    γ_obs_total = [g_obs[g_masks[i]] for i, g_obs in enumerate(γ_obs)]
+    γ_obs_total_censored = [g_obs_cen[g_masks[i]] for i, g_obs_cen in enumerate(gamma_obs_censor)]
+    gammas_total = [np.vstack(gamma_tot)[g_masks[i]] for i, gamma_tot in enumerate(gammas_list)]
+    output = gamma_estimator_atonce(γ_obs_total, γ_obs_total_censored, gammas_total)
```
aarmey, line 222: @Farnazmdi why would running the function twice help to speed it up?

Farnazmdi: I know sorry, I am working on it in local, it should have been recursive, not running that twice. But the problem is that when I call the function inside it, it doesn't get out of it and just keeps running itself. I should find a good stopping point.

aarmey: Why would making it recursive help in any way?

aarmey: When I said "last iteration", I meant last iteration of Baum-Welch, not of the optimization here. Think through the problem. There is also the separated estimator as an example.

Farnazmdi: It should be fixed now. I appreciate if you could review it.

aarmey (commented):
This looks right! Let me review it once more tomorrow morning before merging.

Farnazmdi:
Thank you.

lineage/BaumWelch.py
```diff
@@ -86,9 +87,10 @@ def do_M_step(tHMMobj, MSD, betas, gammas):
 
     if tHMMobj[0].estimate.fE is None:
         assert tHMMobj[0].fE is None
-        for idx, tt in enumerate(tHMMobj):
-            do_M_E_step(tt, gammas[idx])
-
+        if len(tHMMobj) == 1: # means it only performs calculation on one condition at a time.
+            do_M_E_step(tHMMobj[0], gammas[0])
+        else: # means it performs the calculations on several concentrations at once.
```
aarmey, line 92: Does this mean this will be applied in Fig. 10? Do the emissions we generate in that figure satisfy our constraint here? We'll also need to adjust the degrees of freedom provided to AIC (not here but make an issue).

lineage/BaumWelch.py
```diff
@@ -144,6 +146,31 @@ def do_M_E_step(tHMMobj, gammas):
...
 
+def do_M_E_step_1state(all_tHMMobj, gammas_1st, state_j):
+    """ perform the M_E step when all the concentrations are given at once for one state.
+    all_cells is a list (for different concentrations) containing lists of tuples of observations. 
+    gammas_1st is a list of lists, each cell's gamma value for one state. """
+    G1cells = []
+    G2cells = []
+    for tHMMobj in all_tHMMobj:
+        all_cells = np.array([cell.obs for lineage in tHMMobj.X for cell in lineage.output_lineage])
+        G1cells.append(all_cells[:, np.array([0, 2, 4])])
```
aarmey, line 157: This is collecting the emissions, right? Given that these don't depend on the state, why not do this in `do_M_E_step_atonce` rather than `do_M_E_step_1state`?

lineage/states/stateCommon.py
```diff
@@ -105,3 +105,44 @@ def basic_censor(cell):
...
+        x0 = np.array([20.0, 2.0, 3.0, 4.0, 5.0])
+
+    def constr(x):
+        return x[2:5] - x[1:4]
+
+    def constr2(x):
+        return 100.0 - x[0]
+
+    def constr3(x):
+        return 20.0 - x[1:5]
```
aarmey, line 145: Is this needed? What was happening when the upper bound was just uniformly 100?

lineage/tHMM.py
```diff
@@ -191,10 +189,9 @@ def fit_list(tHMMobj_list, tolerance=1e-9, max_iter=1000):
     # first stopping condition check
     for _ in range(max_iter):
         do_M_step(tHMMobj_list, MSD_list, betas_list, gammas_list)
-
         MSD_list, NF_list, betas_list, gammas_list = map(list, zip(*[do_E_step(tHMM) for tHMM in tHMMobj_list]))
         new_LL = np.sum([np.sum(calculate_log_likelihood(NF)) for NF in NF_list])
-
+        print("new LL", new_LL)
```
aarmey, line 194: Remove printing.

lineage/tHMM.py
```diff
@@ -173,16 +169,18 @@ def log_E_score(EL_array, state_tree_sequence):
     return log_E_score_holder
 
 
-def fit_list(tHMMobj_list, tolerance=1e-9, max_iter=1000):
+def fit_list(tHMMobj_list, tolerance=1e-4, max_iter=1000):
```
aarmey, line 172: Remove tolerance change.

lineage/states/stateCommon.py
```diff
@@ -105,3 +105,44 @@ def basic_censor(cell):
...
+
+    res = fmin_cobyla(func=negative_LL_atonce, x0=x0, cons=[constr, constr2, constr3], args=arrgs, consargs=(), maxfun=500000, disp=3)
```
aarmey, line 147:
```suggestion
    res = fmin_cobyla(func=negative_LL_atonce, x0=x0, cons=[constr, constr2, constr3], args=arrgs, consargs=(), maxfun=500000, rhoend=1e-6, rhobeg=2.0)
```

Remove printing, lower final tolerance, and allow slightly larger steps at the beginning.

aarmey (changes requested):
Just some small changes, then this looks ready!

lineage/BaumWelch.py
Farnazmdi: Yes, I hope it can help with solving the issue that we had with figure 10 (every time it looked different).
I will make the issue right now.

Farnazmdi: I have to look into the emissions, but I think the data is the same, so should be the emissions.

lineage/BaumWelch.py
Farnazmdi: I removed `do_M_E_step_1state` and all is included in the `do_M_E_step_atonce` now.

lineage/states/stateCommon.py
Farnazmdi: Maybe I'm wrong, but I expect the shape parameter to be larger than the scale parameter.
Given that `shape x scale` should indicate the average of phase lengths, I thought we should restrict them a bit more because their multiplication should not be a very large number.

lineage/BaumWelch.py
Farnazmdi: I will take a closer look at this in another PR.

lineage/states/stateCommon.py
Farnazmdi: Should I bring 800 constraint back?

aarmey: I think just 100 upper bound across the board? Mainly I think it should be consistent with the bounds in the separate estimator.

Farnazmdi: Of course. Will do.

aarmey (approved):
Ready to merge if it all builds fine.

PR #701, opened 2020-11-14
manuscript edits

jclagarde:
These are the edits to the manuscript that should answer #3, 4, and 7 on the list of edits to make. Please let me know if there are any changes I should make. No rush on the feedback, feel free to take a look whenever it's convenient for you. 

I was also thinking that for figure 1, Dr. Gross' comment about plotting the number of cells over time for both subfigure (a) and (b) would make the meaning of figure 1 more clear. What do you think @Farnazmdi ?


Farnazmdi (approved):

Farnazmdi:
Looks good! Thanks.

PR #702, opened 2020-11-16
Estimator

Farnazmdi:
This only includes the code for the estimator for 4 concentrations of a drug at once `[shape_shared], scale1, scale2, scale3, scale4`.


Farnazmdi:
@aarmey The estimator for estimating [shared_shape, scale1, scale2, scale3, scale4] sometimes works and sometimes returns the initial value that has been given. 
Based on this: [https://stackoverflow.com/questions/47004736/scipy-minimize-returns-initial-value](https://stackoverflow.com/questions/47004736/scipy-minimize-returns-initial-value) the function is probably not smooth, but the function is the same and I just added the negative likelihood of different concentration conditions together. 
The only thing left would be the constraints. Right now this is the constraints:
IMG-8588
Am I doing this wrong? Or you think the constraints make it unsmooth?

lineage/states/stateCommon.py
```diff
@@ -105,3 +102,53 @@ def basic_censor(cell):
...
+
+    # A is a matrix of coefficients of the constraints.
+    # For example if we have x_1 - 2x_2 >= 0 then it forms a row in the A matrix as: [1, -2], and one indice in the b array [0].
+    # the row array of independent variables are assumed to be [shape, scale1, scale2, scale3, scal4]
+    A = np.array([[0, 1, -1, 0, 0], [0, 0, 1, -1, 0], [0, 0, 0, 1, -1]])
+    bnds = Bounds([1, 0.001, 0.001, 0.001, 0.001], [100, 50.0, 50.0, 50.0, 50.0]) # list [min], [max]
+    cons = LinearConstraint(A, lb=[-np.inf]*A.shape[0], ub=[0.0]*A.shape[0])
+    res = minimize(fun=negative_LL_atonce, jac="3-point", x0=[10.0, 0.05, 0.05, 0.05, 0.05], bounds=bnds, constraints=cons, args=arrgs, options=opt)
+    xOut = res.x
+
```
Farnazmdi, line 153: I was referring to these lines.

aarmey:
This might not be convex but should be smooth. Have you looked at the verbose output from minimize? It usually provides a bit more information about why it returned. I'm guessing it's a failed line search, in which case we can try a different optimization method.

Farnazmdi:
Thanks very much! I changed it to COBYLA and it worked. (at least didn't fail for a couple of times that I ran the test and I printed the negative likelihood and it seems decreasing nicely.)

lineage/states/stateCommon.py
```diff
@@ -103,3 +103,52 @@ def basic_censor(cell):
                 cell.get_sister().left.observed = False
                 cell.get_sister().right.observed = False
 
+def negative_LL_atonce(x, uncens_obs, uncens_gammas, cens_obs, cens_gammas):
+    """ uses the negative_LL_atonce and passes the vector of scales and the shared shape parameter. """
+    return negative_LL_sep_atonce(x[1:5], x[0], uncens_obs, uncens_gammas, cens_obs, cens_gammas)
+
+def negative_LL_sep_atonce(scales, a, uncens_obs, uncens_gammas, cens_obs, cens_gammas):
```
aarmey, line 110: Can this not call `negative_LL_sep` to reuse code better?

lineage/states/StateDistributionGamma.py
```diff
@@ -189,3 +189,24 @@ def time_censor(cell, desired_experiment_time):
...
+    bern_obs = [x[:, 0] for x in x_data]
+    γ_obs = [x[:, 1] for x in x_data]
+    gamma_obs_censor = [x[:, 2] for x in x_data]
+
+    # Both unoberved and dead cells should be removed from gamma
+    g_masks = [np.logical_and(np.isfinite(γ_o), berns) for γ_o, berns in zip(γ_obs, bern_obs)]
+    for g_mask in g_masks:
+        assert np.sum(g_mask) > 0, f"All the cells are eliminated from the Gamma estimator."
+
+    γ_obs_total = [g_obs[g_masks[i]] for i, g_obs in enumerate(γ_obs)]
```
aarmey, line 208: Rewrote these as generators instead of loops for clarity.

lineage/states/stateCommon.py
```diff
@@ -103,3 +103,52 @@ def basic_censor(cell):
...
+            gammas.append(gamma)
+
+    for i, gamma in enumerate(gammas):
+        assert gamma.shape[0] == gamma_obs[i].shape[0]
+    arg1 = []
+    arg2 = []
+    arg3 = []
+    arg4 = []
+    for i in range(len(gamma_obs)):
+        arg1.append(gamma_obs[i][time_cen[i] == 1])
```
aarmey, line 138: This would be better as a generator than as loops. `arg1 = [gamma_obs[i][time_cen[i] == 1] for i in range(len(gamma_obs))]` as an example.

aarmey (approved):
Great. Check that you're getting the expected results. I also made a couple comments about minor code changes. After that this looks ready to merge.

Farnazmdi:
Thank you very much. I will edit based on the comments. 
The build errored last time and it was this:
image
I didn't find anything online about this error. Do you have an idea of what should be fixed?

aarmey:
I think I just fixed that with changes in master.

lineage/states/stateCommon.py
Farnazmdi: done.

Farnazmdi: I had to return the code back because of a dimension mismatch that I had to fix in the `negative_LL_sep_atonce()`.

aarmey: Ok, we can always simplify the code later.

PR #706, opened 2020-11-28
Bw

Farnazmdi:
This PR includes code to link the estimator to the group of data we want to fit at the same time.  
@aarmey This is the PR that doesn't update negative likelihood. 
We have a `M_step_for1state` and another one that runs this for several steps. I printed the negative likelihood, but it doesn't change and only runs everything once.


Farnazmdi (commented):
Just to clarify for myself; I should edit the E step (and everything in it) so that it collects all the cells and performs MSD, NF,... at once, right? @aarmey

aarmey:
I don't think you should have to modify the E step at all.

aarmey (changes requested):
@Farnazmdi I fixed the optimization here. COBYLA expected the constraints to be in some weird other format, and it was immediately exiting without an error. The fitting now occurs over multiple iterations.

The figure plots still don't look quite right, and I think it has something to do with the scales ending up really small. Take a look at this and let me know what you find.

Farnazmdi:
Thank you so much, I will look into it.

Farnazmdi:
Findings:
I realized that the likelihood still doesn't update. The EM runs only three times and among these three, only one of them is in the fol loop that really counts. I mean the `new_LL` and `old_LL` in the `/tHMM/fit_list()`.
I also printed the output of `negative_LL_atonce()` I mean the `outt` variable. It stays zero every time it is being called. I probably have not connected a piece of the whole system to the `at_once` parts. My guess is that the `gammas` are not being updated.

Farnazmdi:
update:
1. The parameters were not assigned correctly for each phase, now it is fixed.
figure 11 now looks more reasonable
image
2. The test is fixed and checks if the constraint conditions are satisfied.

The problems:
1. The optimization _sometimes_ fails because the likelihood becomes NaN. I still need to figure out where that may come from.
2. The EM still runs once, i.e., the negative likelihood in the `fit_list` still doesn't change from one iteration to another, and figure 11 above is the result of only running the EM once. The Likelihood is calculated from the `NF`. I looked almost half a day yesterday trying to figure out what is wrong with that, but no success yet.

aarmey:
Does NF and gamma change from the first to second iteration?

aarmey:
Also, it looks like all the cells are being assigned to one state. I'd check that you're putting the optimization results back into the tHMM correctly for all of the states.

Farnazmdi:
The NF doesn't but the gamma changes from the first step to the second.
I will check that, thank you.

aarmey:
So whatever is between those two is the problem. Betas is calculated between those two, so check that. Then, I *think* the difference between gammas and NF/betas is the emissions, so that would possibly suggest the emissions distributions aren't properly updated.

Farnazmdi:
I will look into those. Thank you.

Farnazmdi:
@aarmey I found the problem, but couldn't figure out the solution yet. The problem is that the parameters only in the `pdf` function are not being updated. [link to the pdf function inside the StateDistributionGamma](https://github.com/meyer-lab/tHMM/blob/30f131dd92f33a13f2e0eb336b09e09e64739d46/lineage/states/StateDistributionGamma.py#L54)
It is calling `self.params` that are from `__init__`. 
Is there a way I can overwrite that?

aarmey:
Yes—you overwrite `X.params` within the class. You can overwrite any class members within python; nothing is restricted.

aarmey:
init is only called when the class is first created, so if you overwrite params after it's created it will be what you set.

Farnazmdi:
Thank you, I didn't know it has to be _inside_ the class. I had a function outside the class that used to overwrite the parameters. They were updated _everywhere_ except for that `pdf` function which sounded weird.

update: I have a method inside the GammaDistribution class now that updates the parameters, but it only takes longer, and the result is the same as before. I will think about it more and update you here.

lineage/BaumWelch.py
```diff
@@ -147,33 +146,31 @@ def do_M_E_step(tHMMobj, gammas):
...
+    G1cells = []
+    G2cells = []
     for tHMMobj in all_tHMMobj:
-        all_cells.append([cell.obs for lineage in tHMMobj.X for cell in lineage.output_lineage])
-        # TODO: the line where we pass the cells and gammas to the atonce_estimator
-        for i in range(len(tHMMobj.estimate.E[state_j].params)):
-            tHMMobj.estimate.E[state_j].params[i] = 0.5
+        all_cells = np.array([cell.obs for lineage in tHMMobj.X for cell in lineage.output_lineage])
+        G1cells.append(all_cells[:, np.array([0, 2, 4])])
+        G2cells.append(all_cells[:, np.array([1, 3, 5])])
```
aarmey, line 158: Run `atonce_estimator` once outside here, then just assign the results to the various emissions by directly writing to `tHMMobj.estimate.E[state_j].params`. There's no need to write to params from within the class.

aarmey: Also, are you sure that `do_M_pi_step` isn't being called when you do the combined fitting? That would overwrite your results.

Farnazmdi: Thank you. I will bring that back.
Why would `do_M_pi_step` overwrite the parameters? I didn't see anything being assigned to the tHMMobj in `do_M_pi_step` or `do_M_T_step`. Could you elaborate please?

aarmey: Oh, I meant `do_M_E_step`.

Farnazmdi: I have a separate `do_M_E_step` that when fitting at once, it is called.
I will check again. Thank you.

Farnazmdi:
Update: The problem was that the parameters of G1 and G2 like `StateDistribution.G1.params` were not being updated. That is fixed now, but the optimizer won't succeed sometimes (the assertion of `assert res.success` fails).
I will work on it today and see if I can find the problem.

aarmey:
Ok, if you print out res and post it here I might have some thoughts about why it's failing.

Farnazmdi:
Sorry to reply late, I was away from my computer.
It seems strange that the negative likelihood is positive numbers and it is decreasing with more iterations.
Sometimes the negative likelihood becomes NaN and that's where the error occurs.

If it doesn't become NaN, the `res.success` fails that in this case this is the `res`:
image

aarmey:
This failure is just because it hit the maximum number of iterations. You can increase `maxiter` in `options` to 10000 or some such.

Farnazmdi:
I increased it to 30000 but still fails.

aarmey:
With the same message?

Farnazmdi:
Yes. It only happens for figure 12. Figure 11 builds fine.

aarmey:
Can you post the `res` output again then?

Farnazmdi:
Absolutely,
image

aarmey:
This still shows the maximum number of iterations as set to 1000?

Farnazmdi:
I set the `max_iter` in the `fit_list` to 30000. You are right I didn't change the iterations in the optimizer. 
I will see how to do that right away.

Farnazmdi:
I increased it, and now every time I get this error regarding the optimizer. 
image
I think the likelihood becomes NaN. Could that be that the pdf is returning nan?

aarmey:
Yep. Check which inputs lead to NaN outputs within the LL function. It's likely something easily controlled with narrower bounds.

Farnazmdi:
I checked, and it seems somewhere the scale parameter is a negative number, but the bounds shouldn't allow it, they are from 0.001 to 500. 
It might be the constraints are not properly defined. I will check that.

aarmey:
Look at `keep_feasible`: https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.Bounds.html#scipy.optimize.Bounds

aarmey:
@Farnazmdi did that option help?

Farnazmdi:
I tried `keep_feasible` but sometimes it errors for the initial value (says "x0 violates bound constraint!" although it doesn't. Maybe there is another bound that I don't see). I reduced the `xtol` and 'gtol' to 1e-4 without using the `keep_feasible` option, but still even for 100000 iterations, it won't be successful.

Farnazmdi:
This is the setting now, does it violate the constraints?
image

aarmey:
Are you *sure* it sometimes says that about x0? Your inputs are the same each time so that would be very surprising.

Farnazmdi:
No sorry, it always happens when I have the `keep_feasible` option.

aarmey:
The your linear constraint isn't what you expect. Have you tried looking at `A @ x0` since that's what it does?

Farnazmdi:
I will do it right now. Thanks.

Farnazmdi:
Update:
I looked at it, the vector `A dot x0` is `[-0.2, -0.2, -0.2]^T` which is greater than lb(-inf) and less that ub (zeros). It doesn't violate the constraints.
I tried changing the optimizer to `cobyla` and at first it wasn't successfully converging because of the number if `maxiter`, I increased it to 1e6 and it doesn't error, but also it is taking a really long time to run.
 This is what it displays every time it runs the minimization:
image
It is just taking a very long time.

Farnazmdi:
@aarmey Could we use parallel computing here? I remember we had it but you said it was slowing down instead of speeding up.

aarmey:
Do the eventual results look right? There’s probably a variety of ways we can speed it up once we’re sure the optimization problem is defined correctly.

Farnazmdi:
Yes! I built the figures.
Here is figure 11 (lapatinib)
image
And here is figure 12 (gemcitabine):
image

Although the build files due to some issue in building the manuscript. Could you please give me a hint on how to fix it?
image

aarmey:
Ok great. Please merge this PR. I’ll fix that build error and look into the performance tomorrow.

PR #707, opened 2020-12-01
Remove duplicate nLL function

aarmey:


Farnazmdi (approved):
Thank you very much. I had a problem with the dimensions that I didn't use `negative_LL_sep` in the `atonce` function. I guess I wasn't familiar with `np.squeeze`.

aarmey:
Yeah, it was an esoteric issue—when there were no observations, one array was size 0, and the other was 1x0. Numpy is strange in that empty arrays can have differing dimensions.

PR #712, opened 2020-12-11
dof and aic fixes

Farnazmdi:


Farnazmdi:
@aarmey I can't find the file we used to set the maximum run time, however, the build is taking more than 6 hours. But I think they are being built (no errors).
Could you give me some clues on how to make it faster?

aarmey:
I assume it’s one figure taking so long? I’ll have to look.

Farnazmdi:
No actually, those figures (9, 11, 12) that are using the fitting atonce are taking ~ 8000 to ~ 12000 seconds

aarmey:
See #713

Farnazmdi:
I was able to build figure 9 in local after 4800 seconds. Although, it sometimes fails due to NaN or Inf in the value of the negative likelihood for data in the high concentrations of gemcitabine. It looks like this:
image

aarmey:
How did you determine the degrees of freedom?

Farnazmdi:
Since the shape parameter is shared between the set of populations that are fitted simultaneously, I assumed each population has 2.25 number of free parameters (bern, scale, 1/4 shape) which makes it 4.5 for the phase-specific distributions. The `pi` and `T` is the same because each time we are evaluating all of the populations for a specific number of states, so I just said it should look at the first tHMMobj (control).

Farnazmdi:
It built! could you please review this?

lineage/Analyze.py
```diff
@@ -31,7 +31,7 @@ def Analyze_list(Population_list, num_states, **kwargs):
     return tHMMobj_list, pred_states_by_lineage_by_conc, LL
 
 
-def run_Analyze_over(list_of_populations, num_states, parallel=True, **kwargs):
+def run_Analyze_over(list_of_populations, num_states, parallel=True, atonce=False, **kwargs):
```
aarmey, line 34: Currently `fit_list` always runs in a combined way so long as two or more conditions are included. Should it see this option?

lineage/states/StateDistributionGamma.py
```diff
@@ -15,6 +15,7 @@ def __init__(self, bern_p=0.9, gamma_a=7, gamma_scale=4.5, shape=None):
         """ Initialization function should take in just in the parameters for the observations that comprise the multivariate random variable emission they expect their data to have. """
         self.params = np.array([bern_p, gamma_a, gamma_scale])
         self.const_shape = shape
+        self.degFreedom = 0
```
aarmey, line 18: Where is this used?

lineage/tHMM.py
```diff
@@ -97,8 +97,12 @@ def get_AIC(self, LL):
         if self.fT is None:
             degrees_of_freedom += self.num_states * (self.num_states - 1)
 
-        for ii in range(self.num_states):
-            degrees_of_freedom += self.estimate.E[ii].dof()
+        if atonce:
+            for ii in range(self.num_states):
+                degrees_of_freedom += 4.5
```
aarmey, line 102: Add a comment here that this assumes you're using the cell cycle model, because otherwise the tHMM code works for any type of emissions.

aarmey (changes requested):

lineage/states/StateDistributionGamma.py
Farnazmdi: Removed.

lineage/tHMM.py
Farnazmdi: done.

lineage/Analyze.py
Farnazmdi: The reason I added this argument was that `run_Analyze_over` was being used to run Analyze, which puts the given population in another square bracket since it is assuming it is not given a list of populations (please see line 10-13). I did this so we call `Analyze_list()` without passing through `Analyze()`. Please let me know if you were referring to another point.

aarmey: Could it just check whether it was given a list, then?

Farnazmdi: done.

lineage/figures/figure9.py
```diff
@@ -19,34 +19,24 @@ def makeFigure():
...
+
+    def find_AIC(data, desired_num_states):
+        # Copy out data to full set
+        dataFull = []
+        for _ in desired_num_states:
+            dataFull.append(data)
+
+        # Run fitting
+        output = run_Analyze_over(dataFull, desired_num_states)
+        AICs = np.array([oo[0][0].get_AIC(oo[2])[0] for oo in output])
```
aarmey, line 33: Doesn't this need `atonce=True`?

lineage/tHMM.py
```diff
@@ -97,8 +97,12 @@ def get_AIC(self, LL):
         if self.fT is None:
             degrees_of_freedom += self.num_states * (self.num_states - 1)
 
-        for ii in range(self.num_states):
-            degrees_of_freedom += self.estimate.E[ii].dof()
+        if atonce: # assuming we are fitting all 4 concentrations at once and we have cell cycle phase specific distributions.
+            for ii in range(self.num_states):
```
aarmey, line 101: Do you need a loop here? Can't you just say `degrees_of_freedom = self.num_states * 4.5`?

lineage/figures/figure9.py
Farnazmdi: I removed that argument and instead of it I used `isinstance(population, list)`. I thought you meant this

aarmey: Yes, but you have an if statement in `get_AIC`.

Farnazmdi: Oh that.. sorry, I'll fix it.

aarmey (approved):

lineage/Analyze.py
Farnazmdi: Doing this made some other problem. I couldn't figure out what is wrong. The error is with line 71 in the Analyze file. It says `lineage object is not subscriptable`. 
image
I tried printing the `output` and the `prom_holder`, but I am not sure what `.result()` does.

aarmey: I think this is because Analyze always expects population to be a list. Put population in a list, `[population]` when you directly call it.

Farnazmdi: But if it should be a list, then the line that chooses whether the population goes to `Analyze_list()` or `Analyze()` would be useless.

lineage/Analyze.py
```diff
@@ -63,7 +63,10 @@ def run_Analyze_over(list_of_populations, num_states, parallel=True, **kwargs):
 
         prom_holder = []
         for idx, population in enumerate(list_of_populations):
```
aarmey, line 65: Oh wait... shouldn't this if statement be outside of this loop?!

Farnazmdi: We mostly give lists to the `run_Analyze_over`.

aarmey: Oh right... lists of different state numbers for example... ok, let me take over this branch. I need to adjust it to see what's wrong.

Farnazmdi: Ok, thank you very much.

lineage/Analyze.py
```diff
@@ -63,7 +63,10 @@ def run_Analyze_over(list_of_populations, num_states, parallel=True, **kwargs):
...
-            prom_holder.append(exe.submit(Analyze, population, num_states[idx], fpi=list_of_fpi[idx], fT=list_of_fT[idx], fE=list_of_fE[idx]))
+            if isinstance(population, list):
```
aarmey, line 66: @Farnazmdi I was wrong—it will be hard to determine what you want to do here based on population. You'll want to bring back your option to `run_Analyze_over` for atonce fitting.

I can't get the build errors to happen. When running the parallel code, the underlying error doesn't get output, so it's hard to say what failed.

aarmey: You can take it from here.

Farnazmdi: I will do that, thank you.

aarmey: Finally, check by putting `assert False` in the right places that you indeed are *only* getting atonce fitting in Fig. 9, and are not running it in another figure where it shouldn't be called.

aarmey:
Yep... it looks like it has been Figs. 12 and 13 failing, and Fig. 9 looks like it errored just because it was still running (and therefore got killed).

PR #713, opened 2020-12-14
Fixed constrained optimization

aarmey:
The COBYLA optimization turned out to be very bad for this problem, because of the high interdependency between shape and scale. It basically works within square regions of parameter space, so it was spending huge numbers of iterations slightly increasing shape and then slightly decreasing all the scales to compensate.

The trust-region solver is much more mature, so that is implemented here. The other problem was that the first `x0` provided was bad (I mentioned this might be the case). All the scales were the same, which was causing some issue in the solver. I have a quick fix here, but really the tHMM should be providing `x0=None` on the first iteration.

### DIFF
--- a/lineage/states/stateCommon.py
+++ b/lineage/states/stateCommon.py
@@ -145,6 +145,6 @@ def gamma_estimator_atonce(gamma_obs, time_cen, gamas, x0=None):
 
     linc = LinearConstraint(A, lb=np.zeros(3), ub=np.ones(3)*100.0, keep_feasible=True)
     bnds = Bounds(lb=np.zeros_like(x0), ub=np.ones_like(x0)*100.0, keep_feasible=True)
-    res = minimize(negative_LL_atonce, x0=x0, args=arrgs, method="trust-constr", hess="3-point", bounds=bnds, constraints=[linc])
+    res = minimize(negative_LL_atonce, x0=x0, args=arrgs, method="trust-constr", hess="HessianUpdateStrategy", bounds=bnds, constraints=[linc])
 
     return res.x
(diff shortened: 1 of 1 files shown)


lineage/states/stateCommon.py
```diff
@@ -135,11 +135,16 @@ def gamma_estimator_atonce(gamma_obs, time_cen, gamas, x0=None):
...
-    def constr(x):
-        return x[2:5] - x[1:4]
+    A = np.zeros((3, 5))
+    np.fill_diagonal(A[:, 1:], -1.0)
+    np.fill_diagonal(A[:, 2:], 1.0)
 
-    def constr2(x):
-        return 100.0 - x
+    # Override x0 if we were given a bad starting point
+    if np.allclose(np.dot(A, x0), 0.0):
```
aarmey, line 143: We might need to also check `x0` before starting, and if it's violating the bounds move it back to a reasonable start.

Farnazmdi: Thank you.
This is the reason I changed it to `cobyla` but you are right, it should be an easy fix.

Farnazmdi (approved):
I thought you left the part related to constraint violation to me, and I found [this](https://github.com/scipy/scipy/issues/11649)  and was trying to change the hessian calculating method, but I think your edit for modifying x0 when it fails, works often. (when I first looked at the PR, the build had failed due to this constraint violation, but I guess it happens less frequently, because I was able to build the two figures with these changes and at another time figure 12 failed.)
I will take a closer look and update you here.

aarmey:
Yes, feel free to adjust this. You can try removing `keep_feasible` perhaps?

PR #717, opened 2020-12-18
Changes

aarmey:
- Make it easier to run Analyze with/without parallelization.
- Remove constant shape methods.
- Lower the upper bound on number of states.

I haven't been able to get this crash. @Farnazmdi, I think you mentioned it happened at the highest number of states, so I'm suggesting we just lower the number of states we evaluate in the AIC figure.

### DIFF
--- a/lineage/figures/figure12.py
+++ b/lineage/figures/figure12.py
@@ -20,15 +20,6 @@ concsValues = ["cntrl", "5nM", "10nM", "30nM"]
 tHMM_solver = tHMM(X=data[0], num_states=1)
 tHMM_solver.fit()
 
-constant_shape = [int(tHMM_solver.estimate.E[0].params[2]), int(tHMM_solver.estimate.E[0].params[4])]
-
-# Set shape
-for population in data:
-    for lin in population:
-        for E in lin.E:
-            E.G1.const_shape = constant_shape[0]
-            E.G2.const_shape = constant_shape[1]
-
 num_states = 3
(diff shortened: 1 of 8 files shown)


lineage/figures/figure12.py
```diff
@@ -20,15 +20,6 @@
 tHMM_solver = tHMM(X=data[0], num_states=1)
 tHMM_solver.fit()
 
-constant_shape = [int(tHMM_solver.estimate.E[0].params[2]), int(tHMM_solver.estimate.E[0].params[4])]
```
aarmey, line 23: This file isn't yet updated to run `atonce`.

Farnazmdi: Thank you, I will fix this in my branch.

Farnazmdi:
Thank you very much! 
I think the problem was with high concentration data, I am not sure if the high number of states makes problems or not, but my PR keeps failing because of that and I haven't merged it.

aarmey:
This is set to merge into your branch.

Either way, evaluating fewer states will significantly speed up the figure build, which will help debugging.

aarmey:
Actually, here figure 12 failed, which isn't even using the `atonce` fitting at the moment...

aarmey:
https://github.com/meyer-lab/tHMM/pull/718/checks?check_run_id=1577850922

Farnazmdi:
Thank you, I will look into figure 12.

PR #727, opened 2021-01-03
Fig1112

Farnazmdi:


Farnazmdi:
@aarmey I saved figure 11 and 12 in ``lineage/figures/cartoons` as `fig11.svg` and `fig12.svg`, but tried to address it as `./lineage/figures/cartoons/figu11.svg` or `./output/fig11.svg` (while copying it to the output folder) but it doesn't show it. I looked at figure 2 which is purely a cartoon, but we are just calling it from the output folder and it appears. Could you please take a look at this?

genFigures.py
```diff
@@ -88,12 +88,12 @@
...
-    if sys.argv[1] == '11':
-        # Overlay Transition block
-        overlayCartoon(fdir + 'figure11.svg',
-                       f'{cartoon_dir}/lapatinib.svg', 930, 120, scalee=0.9)
-
-    if sys.argv[1] == '12':
-        # Overlay Transition block
-        overlayCartoon(fdir + 'figure12.svg',
-                       f'{cartoon_dir}/gemcitabine.svg', 930, 120, scalee=0.9)
+    # if sys.argv[1] == '11':
```
aarmey, line 91: @Farnazmdi I don't think you want to comment these out.

Farnazmdi: Oh.. I will revert it. Thanks.

PR #733, opened 2021-01-14
fig 3 fix

Farnazmdi:
I have been trying to fit censored data when the model doesn't know the data is censored (to create the suggested part (c)) but it fails to run. There is an assertion error for estimating the transition probability matrix. 
So, instead of keeping the censored observations and treating them like observed observations, I marked them as 'nan' so it seems we removed the censored observations. 
Do you expect it to be this bad? @aarmey 

image


aarmey:
But this doesn't match the process we're trying to imitate.

What is the assertion error? You should fix that.

Farnazmdi:
By "uncensored model" you mean the model thinks of censored data as uncensored, right? 

Sometimes the `res.success` from the gamma_estimator fails, sometimes `np.all(np.isfinite(T_estimate))` from M_T step fails.

aarmey:
Censored is defined by the Bernoulli being NaN, right?

If you print out `res` when it fails that check, what does it show?

aarmey:
You might have to mark all the censored cells at the end of the experiment with a non-NaN Bernoulli.

Farnazmdi:
I was focusing on the 4th and 5th elements of the `cell.obs`. If cell.obs[4] == 0 it means G1 is censored, if == 1 means uncensored, and if 'nan' means unobserved. The 5th element is the same for G2 phase.

aarmey:
In the "broken" case, you want the model to think these are fully observed cells, even though they aren't. So, they should be marked with 1 as uncensored.

Farnazmdi:
I was marking cell.obs[4] == cell.obs[5] == 1 so that the model thinks they are uncensored. But never got a successful run.

Farnazmdi:
This is the message when the convergence is not successful:
image

aarmey:
Ok, you can comment out that convergence assert. Something weird is going on there, but it’s close to the solution.

Farnazmdi:
The problem with other assertion errors originates from pdf calculation for some of the cells.
When a pdf is 'nan', how is it different from pdf being 0?

aarmey:
Hmm... do you know the inputs that result in NaN? It shouldn't happen, and is not the same as 0.

https://en.wikipedia.org/wiki/IEEE_754-1985

Farnazmdi:
There are many cells. I will try to print out their observations tomorrow.

Farnazmdi:
image
I had to leave the unobserved cells to be marked as `nan`. I think it is fixed now.

lineage/states/stateCommon.py
```diff
@@ -68,15 +68,14 @@ def gamma_estimator(gamma_obs, time_cen, gammas, x0):
     assert gammas.shape[0] == gamma_obs.shape[0]
     arrgs = (gamma_obs[time_cen == 1], gammas[time_cen == 1], gamma_obs[time_cen == 0], gammas[time_cen == 0])
     opt = {'gtol': 1e-12, 'ftol': 1e-12}
-    bnd = (1.0, 800.0)
+    bnd = (1.0, 100.0)
```
aarmey, line 71: I assume this bound is still sufficiently wide to cover all feasible values?

aarmey (approved):
One question—depending upon the answer there this is ready.

lineage/states/stateCommon.py
Farnazmdi: Oh I tried this to fix the error in the tests. I was passing almost 99% when I changed it to 100. I'm not sure if this was the problem though. Should I bring that back?

Farnazmdi: To answer your question, I think it is.

aarmey: Ok, this is good to merge.

PR #735, opened 2021-01-18
figure 9 update

Farnazmdi:
image
I removed the check for the success of the optimization. It wouldn't become successful even for 100000 iterations. I wasn't sure what is best to do, I let it go as far as 1000 iterations. Please let me know if you think I should change the iterations or something else.

In this branch:
1. I ran figure 9, ran and pickled the model on both lapatinib and gemcitabine. 
 2. Edited code in figure 10, now figures 10, 101, 102, and 103 show single cell trees with their states for control, 25 nM, 50 nM, and 250 nM for lapatinib, respectively. I made them separately because I tried putting all 4 conditions together but it took forever to build (I let it run from 9 pm yesterday and at 6 am today it was still running)
3. Made the same sort of figures for gemcitabine with names figure 151, 152, and 153 showing 5, 10, and 30 nM gemcitabine.
** All the new tree figures use the pickled analysis, so they are all consistent.

I have not placed them in the manuscript yet. I would appreciate your opinion on where they should go. They are in the output file right now.


Farnazmdi:
Done.
The tree figures, however, take almost 500 seconds each. I think the total build time would be ~ 25000 seconds.

aarmey:
That's strange. Have you profiled to see what they're spending time doing?

Farnazmdi:
I found that strange, too. Because running the function that plots each tree doesn't take more than 15 seconds for all of the times that it has to be run. Most of the time is taken after the line `return f` at the very end of `makeFigure()`. It looks like loading all these graphics into a file takes time.

aarmey:
Hmm... something must be really inefficient in the phylogeny package. Ok, include the svg files then... you can just revert your recent commits.

Farnazmdi:
Should I include these tree plots in the main figures of manuscript?

aarmey:
What do you think? What do they show?

Farnazmdi:
I remember we discussed they should be in the manuscript to show the type of data we used. These show the state of the cells, so I guess if we only want to show them as the type of data, they shouldn't be colored.
On the other hand, we could bring them at the end of the results, maybe beside their corresponding drug results and show what we mean by assigning states to every single cell.

There is also benefit in bringing both. I mean, having a sample of the data but without coloring them at the beginning of the results, maybe along with figure 2, and also bringing a sample of them, after coloring, at the end along with the analysis for their drug.

Farnazmdi:
If it is okay, we merge this and I'll figure out where these should go in the manuscript.

aarmey (approved):

PR #736, opened 2021-01-19
Update 05.methods.md

Farnazmdi:
I think I had forgotten to *add* the free parameters to the degrees of freedom. Does this make sense now?


aarmey (approved):
Yep! Good catch.

PR #738, opened 2021-01-20
resized figure 6

jclagarde:
Hello! I just resized the cartoon in figure 6, which is also titled figure 5 in the manuscript, so that it can better fit the empty space in the figure. I cloned the repo so that my local tHMM files should be up to date but please let me know if it's still having issues


Farnazmdi:
Hi JC,
I actually had resized it because it seemed a little too large.
The AIC is figure 6. I guess they are titled correctly.

jclagarde:
@Farnazmdi ah okay, sounds good!

PR #747, opened 2021-02-06
bounds and lineages

Farnazmdi:
1. added the arguments `keep_feasible` for bounds in stateCommon.
2. fixed the importing problem and divided the phase lengths by 2.


lineage/data/Lineage_collections.py
```diff
@@ -28,7 +23,8 @@ def popout_single_lineages(lineages):
 lap02 = [LineageTree(list_of_cells, E) for list_of_cells in import_Heiser(path=r"lineage/data/heiser_data/new_version/AU00701_A5_1_V4.xlsx")]
 lap03 = [LineageTree(list_of_cells, E) for list_of_cells in import_Heiser(path=r"lineage/data/heiser_data/new_version/AU00801_A5_1_V4.xlsx")]
 
-Lapatinib_Control = lap01 + lap012 + lap02 + lap03
+lapatinib_Control = lap01 + lap012 + lap02 + lap03
+Lapatinib_Control = fix_obs(lapatinib_Control)
```
aarmey, line 27: Why is this being done over and over after import? Shouldn't the fix be on import since it affects every file?

Farnazmdi: fixed.

aarmey:
I changed how the Hessian is calculated. Let me know if this new way works.

Farnazmdi:
I still get the error.

Farnazmdi:
wait, I am getting some error from the importing test. I will check that, it might be the problem.

lineage/data/Lineage_collections.py
```diff
@@ -106,8 +94,7 @@ def popout_single_lineages(lineages):
 gem06 = [LineageTree(list_of_cells, E) for list_of_cells in import_Heiser(path=r"lineage/data/heiser_data/new_version/AU00801_A3_1_V4.xlsx")]
 gem062 = [LineageTree(list_of_cells, E) for list_of_cells in import_Heiser(path=r"lineage/data/heiser_data/new_version/AU00801_A3_2_V4.xlsx")]
 
-gemControl = gem04 + gem05 + gem06 + gem042 + gem052 + gem062 + gem043
-Gemcitabine_Control = popout_single_lineages(gemControl)
```
aarmey, line 110: Why did you remove `popout_single_lineages`? I'd guess that's why this is failing.

Farnazmdi:
The problem with the test was irrelevant. Should I try and implement the Hessian? 
I commented on one of the issues on their repo and asked about this problem. I hope they get back to me soon.

lineage/data/Lineage_collections.py
Farnazmdi: I tried bringing it back, still got the error.
I thought removing single lineages means removing those that have phase extension. I can remove them if it doesn't create some bias.

aarmey:
I already made the Hessian edit. It still fails? The issue I see here is not fitting.

aarmey:
Oh wait I see the fitting issue.

I don’t understand your other code edits here. Why do you have to remove the test?

lineage/data/Lineage_collections.py
Farnazmdi: When I was debugging the test I realized that we have some cases like:
[nan, nan, 62.5, nan, 0, nan]
which means we don't know about G1 and G2 fates,
the G1 is censored and the length of it is 62.5
we have not observed G2.

Is this something feasible? I mean the experiment time is 96 hours, and the cell is censored in G1, and we don't know about its fate. I mean in this case, shouldn't we have 96 instead of 62.5?

Farnazmdi:
I merged your changes regarding the hessian and it failed again.

I removed that part of the test because it was checking the equality of the observations when we call the `import_Heiser()` versus the observations themselves. Because I had divided the phase lengths by 2, they were not equal. So I removed that part.

aarmey:
The starting point was too far off after scaling, and caused nans on first evaluation. Should be fixed.

aarmey:
Ergh still failing...

Farnazmdi:
I pulled the changes and am running it locally. It has not failed.

lineage/data/Lineage_collections.py
aarmey: And this is a root cell? If so, yes, something is off about this. If you can trace it back to a specific file it'd be helpful to see what's there. What does the 0 represent?

lineage/states/stateCommon.py
```diff
@@ -108,14 +108,22 @@ def basic_censor(cell):
 
 def nLL_atonce(x, uncens_obs, uncens_gammas, cens_obs, cens_gammas):
     """ uses the nLL_atonce and passes the vector of scales and the shared shape parameter. """
+    x = np.clip(x, 0.001, 100.0) # Horrible hack
```
aarmey, line 111: If we're out of the bounds just solve for the closest point in bounds.

lineage/data/Lineage_collections.py
```diff
@@ -185,16 +171,13 @@ def popout_single_lineages(lineages):
 taxb42 = [LineageTree(list_of_cells, E) for list_of_cells in import_Heiser(path=r"lineage/data/heiser_data/new_version/AU00802_B4_1_V4.xlsx")]
 taxb422 = [LineageTree(list_of_cells, E) for list_of_cells in import_Heiser(path=r"lineage/data/heiser_data/new_version/AU00802_B4_2_V4.xlsx")]
 
-taxs = taxb40 + taxb402 + taxb41 + taxb412 + taxb42 + taxb422
-Tax2uM = popout_single_lineages(taxs)
```
aarmey, line 189: Should `Tax2uM` be assembled, just not calling `popout_single_lineages`?

aarmey (approved):
Ok, kind of a terrible hack here, but this now works.

lineage/data/Lineage_collections.py
Farnazmdi: fixed.

Farnazmdi:
Thank you so much. I am pretty sure I had tried different initial vectors, but those I tried hadn't worked.
I will prepare the manuscript with all the figures for tomorrow's meeting and prepare a summary of our plans for ODE. If I could finish it by tonight will send it, if not I will send it tomorrow morning.

PR #750, opened 2021-02-08
Bic c

Farnazmdi:
@aarmey I realized something. These are the parameters for the 4 conditions of lapatinib, for 3 states.
The parameters for states1, 2, and 3 are different across conditions. I thought they should be the same.
I will look and see where should be changed tomorrow. 

image


aarmey:
Yep—looks like only G1 is set correctly.

Farnazmdi:
Sorry I labeled them incorrectly. Shape G2 and scale G1 have to change places.
But I meant among different conditions. Doesn't state 1 have to be the same in all conditions?

aarmey:
Farnaz, at this point you _need_ to be familiar with these concepts from the model. If these are different lineages in the same condition, yes, they should be identical.

Farnazmdi:
Population number 0 is control, population number 1 is treatment with 25 nM lapatinib, ...
 
I think the parameters of state 1 in all conditions should be the same, but they are not here, only the shape is. While the shape has to be shared among all states and all conditions. I am suggesting that we do this modification. Because right now, state 1 has a slightly different meaning in control condition than in 25 nM treatment than in other conditions.

aarmey:
Shape is shared across treatment conditions, the other parameters are not. If all the parameters were shared across the conditions, there would be no way for a drug to have any effect.

Farnazmdi:
Understood. Sorry, it was a very bad mistake. I am very tired and not thinking clearly.

Farnazmdi:
I know that figure 9 ylabel says AIC. Here I replaced AIC with BIC and changed the text accordingly. I will run the edited figure 9 tonight and will update it in the morning. 
Other than this, it should be ready to merge. All figures are up to date and some parts of the text are modified.

lineage/figures/figure9.py
```diff
@@ -20,59 +20,58 @@ def makeFigure():
...
         # Run fitting
         output = run_Analyze_over(dataFull, desired_num_states, atonce=True)
-        AICs = np.array([oo[0][0].get_AIC(oo[2], atonce=True)[0] for oo in output])
+        AICs = np.array([oo[0][0].get_AIC(oo[2], num_cells, atonce=True)[0] for oo in output])
 
         return AICs - np.min(AICs, axis=0)
 
-    lapAIC = find_AIC(lapatinib, desired_num_states)
-    gemAIC = find_AIC(gemcitabine, desired_num_states)
+    lapAIC = find_AIC(lapatinib, desired_num_states, num_cells=5290)
```
aarmey, line 35: What is the numcells argument here for?

lineage/figures/figure9.py
```diff
@@ -20,59 +20,58 @@ def makeFigure():
...
-    lpt_st = desired_num_states[np.argmin(lapAIC)]
-    gmc_st = desired_num_states[np.argmin(gemAIC)]
+    # # Lapatinib
+    # lapt_tHMMobj_list, lapt_states_list, _ = Analyze_list(lapatinib, 3, fpi=True)
 
-    # run analysis for the found number if states
-    lapt_tHMMobj_list, lapt_states_list, _ = Analyze_list(lapatinib, lpt_st, fpi=True)
-    gemc_tHMMobj_list, gemc_states_list, _ = Analyze_list(gemcitabine, gmc_st, fpi=True)
+    # # assign the predicted states to each cell
+    # for idx, lapt_tHMMobj in enumerate(lapt_tHMMobj_list):
```
aarmey, line 42: Is this commented code still needed?

aarmey (changes requested):
Just two questions.

lineage/figures/figure9.py
Farnazmdi: The BIC needs the number of data points, I printed out the number of cells in both drug treatments and placed it here. I didn't change the name of the function, but it is now based on BIC.

lineage/figures/figure9.py
Farnazmdi: Not now because we have the pickled files, but if for any reason we need to run the analysis again, we will. Should I remove it?

aarmey (approved):
Ok, this is good as-is.

PR #778, opened 2021-03-06
Remove synthetic?

aarmey:
Is this argument ever really used? It doesn't seem to break anything when I remove it...

### DIFF
--- a/lineage/LineageTree.py
+++ b/lineage/LineageTree.py
@@ -15,13 +15,12 @@ class LineageTree:
     The lineage tree is then censord based on the censor condition.
     """
 
-    def __init__(self, list_of_cells, E, synthetic=False):
+    def __init__(self, list_of_cells, E):
         self.E = E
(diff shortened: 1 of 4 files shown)


lineage/LineageTree.py
```diff
@@ -15,13 +15,12 @@ class LineageTree:
     The lineage tree is then censord based on the censor condition.
     """
 
-    def __init__(self, list_of_cells, E, synthetic=False):
+    def __init__(self, list_of_cells, E):
         self.E = E
         self.output_lineage = sorted(list_of_cells, key=operator.attrgetter("gen"))
         self.output_max_gen, self.output_list_of_gens = max_gen(self.output_lineage)
-        if not synthetic:
```
aarmey, line 22: This seems to be the only place where the value was used.

Farnazmdi (approved):
I agree, `LineageTree()` is used when dealing with experimental data
and `init_from_parameters()` is used when dealing with synthetic. I think removing it won't change anything.

PR #785, opened 2021-03-08
Time

Farnazmdi:
Figure S15 shows the abundance of each state across time for all conditions.
image

Is this good to go for supplements?


aarmey:
Neat! Yes, let's put this in the supplement.

Why is it flat in the beginning and drops off at the end? Are you counting censored cells (you should be, if not)?

Farnazmdi:
Ah.. you are right, I will edit it and update you. Thanks.

aarmey (approved):
Excellent! What do you think about showing this as a percentage rather than absolute numbers of cells? I'm wondering if that will make relative population changes easier to see...

PR #797, opened 2021-04-05
figures 11 and 12

Farnazmdi:
I did what we talked about Friday regarding figures 11 and 12. I believe if I could open these two files in affinity designer (right now it says doesn't support the type although it is .svg) I can manipulate the unorganized arrows in the transition block manually. I have given 3 subplot spaces in both rows for the transition block and I guess the way it draws it is random, so there is no easy way to fix their positions through code.
image
image

Figure 9 is edited. please let me know if it is good or not.
image

In addition, figure 1 now has 10 lineages for each condition and they are selected randomly.
I will work on the text in the other branch that I have and am addressing the comments if it is okay.
image


aarmey (commented):
Excellent! Only question is, on the x-axis of the lineages, didn't Sean correct us and point out they go from 0 to 96 hrs rather than 0 to 192, because the cells are imaged every 30 mins? How'd this fix end up getting reverted?

aarmey:
Oh... looks like we never fixed that.

Farnazmdi:
I had totally forgotten that. 
sorry could you elaborate on what is not getting reverted? I am not certain what you are referring to.

aarmey:
We never fixed that issue. I thought it had been fixed then came back.

Farnazmdi:
Oh okay. I will fix it today.

Farnazmdi:
image

This is also applied in Figures 11 and 12.

aarmey (changes requested):
Why do only three colors show up in (a) through (d) of Figs. 11 and 12?

Farnazmdi:
I wanted to do the remaining edits in another PR, but it makes sense to do it here, so I updated those lineage tree figures in Figures 11 and 12, as well as S11 and S12 that show more trees. Also did a bit of cleaning in the process.

aarmey:
You didn't change the sag files for 11 and 12.

Farnazmdi:
Apologies...

aarmey (approved):
Great!

Farnazmdi:
I can't merge. Could you please?

PR #808, opened 2021-04-20
changes for the first issue

Farnazmdi:
- removed figure S13
- xaxis added to S11 and S12
- title capitalization fixed.

I will address the other issue in another branch as I have to run all the figures with starting ylim other than 0 and for synthetic data figures takes long.


aarmey (approved):
Thanks.

PR #810, opened 2021-04-20
figures updated

Farnazmdi:
The figures are updated based on the edit to the limits and now they all start from 0.


aarmey (changes requested):
Should gamma k and theta extend to 0?

Farnazmdi:
I think they were already starting at 0. I don't think it matters. Should I make any changes to it?

aarmey:
No, the y-axis in the figures clearly doesn't extend to 0. Yes, I think it would be helpful to do this.

Farnazmdi:
The gamma parameters in supp figures now are in the range of 0 and 20.

aarmey (approved):

PR #812, opened 2021-04-21
reference to methods

Farnazmdi:
This is the very last comment to address. To explain how distinct the states are in the BIC figure for synthetic data, we could refer to the table of parameters. I search and it seemed this is the correct way of referring to another section in markdown. Please let me know if you agree.


aarmey:
I'm not sure that a table reference is needed—obviously this information should be in the methods. I'd rather just ignore this suggestion.

PR #815, opened 2021-04-23
gemcitabine ref

Farnazmdi:


Farnazmdi:
@aarmey do you mind merging this? I replaced the doi by pmid but I am not sure If the problem is resolved. If it still was ??? I will add it to the references manually.

PR #820, opened 2021-04-30
Fig1

Farnazmdi:
Figure 1:
image

I decided to put two hand-picked sub-populations from control (control 1 and control 2) to show intra-population heterogeneity, and plotted a set of random lineages from gemcitabine treatment.
The reason is that when we treat with a drug, the growth slows down and it is hard to find (and to convince) that in control and in gemcitabine treated populations, we have identical initial and end point measurements.
Please let me know what you think.

Figure 5 as shown in slack with minor edits:
 
image

I have another branch for text edits.


aarmey (approved):
Great. Note there are manuscript text changes needed to reflect this figure change.

Farnazmdi:
Absolutely.

PR #829, opened 2021-06-04
log

Farnazmdi:
loge to log10 in figure 8 and 9.


aarmey (changes requested):
This needs to be indicated in the axis label and figure caption.

PR #836, opened 2021-06-28
Lh edits

Farnazmdi:
A few things I was not sure about:
1. Changing "concentration" to "dose".
2. generating an error bar for the BIC figure for lapatinib and gemcitabine.
3. Removing some of the edges with values below some threshold (eg, .05) in lapatinib and gemcitabine transition probability graph? this would simplify the fig to make more easily interpretable.
4. The edit in the discussion may need re-wording.


Farnazmdi:
Sorry.. at first I didn't realize it is an old branch.. Should have made a new one to avoid combining this with the previous merge..

aarmey (approved):
Great.

PR #840, opened 2021-07-24
Edit figure2.svg

Farnazmdi:
in the middle part, I had forgotten to draw the arrows.. now it is corrected.
image


Farnazmdi:
Also, I don't know how important it is, but some part of the URL in the data availability is on the next page. If we could move the whole material and methods to the next page it would be great.

PR #848, opened 2021-09-30
Docstrings

jclagarde:
Hi Farnaz! When you get the chance, could you take a look at the docstrings for Analyze, CellVar, and Lineage tree that I wrote? If there are any errors or edits I should make, please let me know and I can change it. Thank you!

### DIFF
--- a/lineage/Analyze.py
+++ b/lineage/Analyze.py
@@ -15,13 +15,43 @@ class DummyExecutor(Executor):
 
 
 def Analyze(X, num_states, **kwargs):
-    """ Runs a tHMM and outputs the tHMM object, state assignments, and likelihood. """
+    """ 
+    Runs a tHMM and outputs the tHMM object, state assignments, and likelihood 
+    after fitting experimental data from :func: `Analyze_list`. 
+    
+    :param X: list of objects (cells) in a lineage
+    :type: list
+    :param num_states: the number of states to identify
+    :type: int
+    :return: tHMM object after fitting
+    :rtype: object
+    :return: state assignments for each cell in the lineage
+    :rtype: object
+    :return: the likelihood of state assignments for each cell 
+    :rtype: array
+    
+    """
     tHMMobj_list, st, LL = Analyze_list([X], num_states, **kwargs)
     return tHMMobj_list[0], st[0], LL
 
 
 def Analyze_list(Population_list, num_states, **kwargs):
-    """ This function runs the analyze for the case when we want to fit the experimental data. (fig 11)"""
+    """ 
+
+    This function runs the analyze for the case when we want to fit the experimental data. (fig 11)
+    
+    :param Population_list: a list of the population of cells that contain a lineage
+    :type: list
+    :param num_states: the number of states to identify
+    :type: int
+    :return: list of tHMM objects
+    :rtype: list
+    :return: list of predicted states for population of cells
+    :rtype: list
+    :return: the likelihood of state assingments for each cell
+    :rtype: array
+
+    """
     tHMMobj_list = [tHMM(X, num_states=num_states, **kwargs) for X in Population_list]  # build the tHMM class with X
     _, _, _, _, LL = fit_list(tHMMobj_list)
 
@@ -54,7 +84,7 @@ def run_Analyze_over(list_of_populations, num_states, parallel=True, atonce=Fals
     :type: list
     :param num_states: An integer number of states to identify (a hyper-parameter of our model)
     :type num_states: Int or list
-    :return: A list of tuples from passing the populations to Analyze_list() or Analyze()
+    :return: A list of tuples from passing the populations to :func:`Analyze_list` or :func:`Analyze`
     :rtype: list
     """
     list_of_fpi = kwargs.get("list_of_fpi", [None] * len(list_of_populations))
@@ -88,6 +118,16 @@ def run_Analyze_over(list_of_populations, num_states, parallel=True, atonce=Fals
 def Results(tHMMobj, pred_states_by_lineage, LL):
     """
     This function calculates several results of fitting a synthetic lineage.
+
+    :param tHMMobj: the tHMM class that has been built
+    :type: object
+    :param pred_states_by_lineage: list of predicted states for the population of cells
+    :type: list
+    :param LL: the likelihood value of cells being assigned a certain state
+    :type: array
+    :return: a dictionary containing metrics for the accuracy of state estimation, Wasserstein distance 
+             and between state distributions
+    :rtype: dictionary
     """
     # Instantiating a dictionary to hold the various metrics of accuracy and scoring for the results of our method
     results_dict = {}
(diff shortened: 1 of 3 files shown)


lineage/Analyze.py
```diff
@@ -15,13 +15,41 @@ def submit(self, fn, *args, **kwargs):
...
+    :type X: list
+    :param num_states: The number of states to identify
+    :type num_states: int
+    :return: The tHMM object after fitting
+    :rtype: object
+    :return: State assignments for each cell in the lineage
+    :rtype: object
+    :return: The likelihood of state assignments for each cell 
+    :rtype: array
+
```
Farnazmdi, line 32: The LL is the log-likelihood and is not an array, it is a `float` and is the summation of LLs of all cells

lineage/Analyze.py
```diff
@@ -15,13 +15,41 @@ def submit(self, fn, *args, **kwargs):
...
+    :type Population_list: list
+    :param num_states: The number of states to identify
+    :type num_states: int
+    :return: List of tHMM objects
+    :rtype: list
+    :return: List of predicted states for population of cells
+    :rtype: list
+    :return: The likelihood of state assingments for each cell
+    :rtype: array
+    """
```
Farnazmdi, line 52: same as above.

lineage/Analyze.py
```diff
@@ -84,7 +115,18 @@ def run_Analyze_over(list_of_populations, num_states, parallel=True, atonce=Fals
...
-    This function calculates several results of fitting a synthetic lineage.
+    This function calculates several results of fitting a synthetic lineage, such as the accuracy
+    of state predictions for cells in a population.
+
+    :param tHMMobj: The tHMM class that has been built
+    :type tHMMobj: object
+    :param pred_states_by_lineage: List of predicted states for the population of cells
+    :type pred_states_by_lineage: list
+    :param LL: The likelihood value of cells being assigned a certain state
+    :type LL: array
```
Farnazmdi, line 126: same as above.

lineage/CellVar.py
```diff
@@ -18,9 +18,17 @@ class CellVar:
...
     def __init__(self, parent, gen, **kwargs):
-        """Instantiates the cell object.
-        Contains memeber variables that identify daughter cells
+        """
+        Instantiates the cell object.
+        Contains member variables that identify daughter cells
         and parent cells. Also contains the state of the cell.
+
+        :param self: Instance of the cell object
+        :type self: object
```
Farnazmdi, line 27: you don't need to define `self` anywhere.

lineage/CellVar.py
```diff
@@ -36,6 +44,13 @@ def divide(self, T):
         """
         Member function that performs division of a cell.
         Equivalent to adding another timestep in a Markov process.
+
+        :param self: Instance of the cell object
+        :type self: object
```
Farnazmdi, line 49: same.

lineage/CellVar.py
```diff
@@ -95,6 +110,11 @@ def isRootParent(self):
     def get_sister(self):
         """
         Member function that gets the sister of the current cell.
+
+        :param self: Instance of the cell object
+        :type self: object
```
Farnazmdi, line 115: same. In these cases I guess you can just define the output, or what an instance does.

Farnazmdi (commented):
Please see the above comments. Other than those, looks good! Thank you

jclagarde:
> Please see the below comments. Other than those, looks good! Thank you

Sounds good. I will make those edits and for similar errors as well. Thank you!

Farnazmdi (approved):
looks good!
Just a minor issue I saw in a few places, some of the definitions start with capital letters, some start with lower case letters. I suggest either all upper case or lower case.

jclagarde:
> looks good! Just a minor issue I saw in a few places, some of the definitions start with capital letters, some start with lower case letters. I suggest either all upper case or lower case.

Thanks! Good catch. It looks like you did lower case for the other docstrings so I will make them all lower case to be consistent.

jclagarde:
> > looks good! Just a minor issue I saw in a few places, some of the definitions start with capital letters, some start with lower case letters. I suggest either all upper case or lower case.
> 
> Thanks! Good catch. It looks like you did lower case for the other docstrings so I will make them all lower case to be consistent.

I believe this last commit should have fixed these. When you get a chance Farnaz, please look it over and let me know if there are any additional changes I should make. Thank you!

Farnazmdi:
Great! looks good to me. Thanks.
Please tag Dr. Meyer for final review/merge.

jclagarde:
@aarmey Hi Dr. Meyer, when you get the chance, please take a look at these changes I made to the docstrings for Analyze, CellVar, and LineageTree. If it looks okay, please merge the PR. Thank you!

aarmey:
@Farnazmdi where is the journal language that requests this?

Farnazmdi:
It says the package has to be well-documented, but does not provide an exact style for documentation.

aarmey:
Ok. This is potentially useful (I need to read through), but what they are looking for is documentation about how other people can use the code. Look at some of the recently accepted software packages. It's about having documentation like an API reference and tutorials, not docstrings everywhere.

Farnazmdi:
Understood. I am planning to work on the Docs folder and write the tutorials.

lineage/Analyze.py
```diff
@@ -15,13 +15,41 @@ def submit(self, fn, *args, **kwargs):
...
 def Analyze(X, num_states, **kwargs):
-    """ Runs a tHMM and outputs the tHMM object, state assignments, and likelihood. """
+    """ 
+    Runs a tHMM and outputs the tHMM object, state assignments, and likelihood 
+    after fitting experimental data from :func: `Analyze_list`. 
+    
+    :param X: list of objects (cells) in a lineage
+    :type X: list
+    :param num_states: the number of states to identify
+    :type num_states: int
```
aarmey, line 25: I'd rather use [Python type hints](https://docs.python.org/3/library/typing.html) which we can [import into Sphinx](https://github.com/agronholm/sphinx-autodoc-typehints). It's cleaner and shorter.

lineage/Analyze.py
```diff
@@ -15,13 +15,41 @@ def submit(self, fn, *args, **kwargs):
...
+    :return: the tHMM object after fitting
+    :rtype: object
```
aarmey, line 27: An `rtype` of object tells us nothing about what is returned. For example, here, this is a tHMM class, which is a specific type.

aarmey: It should _almost always_ be possible to say something more specific than object.

aarmey (changes requested):

lineage/Analyze.py
Farnazmdi: Thank you, this is certainly more concise. 
@jclagarde could you edit based on this? I think we only keep the :param ...: and remove the :type ...: and include the type inside the () of function arguments.
Please refer to the example in the above website to see how to write the type of output (def(): ->...)
I will do this in my branch, too.

jclagarde: Hello, I'm sorry for the late response -- I was in Big Bear this weekend for a club retreat. But yes, I could take a look at this and make the changes based on these suggestions. Thank you!

jclagarde:
@Farnazmdi I had pushing a commit called "testing" just to try out the new Python type hints on the first two functions of Analyze.py. When you get the chance, could you take a look at it to see if this is what you meant before I change the rest? Thank you!

Farnazmdi:
Hi JC, yes, that is what we meant. Thank you.

PR #849, opened 2021-10-05
cellvar working?

Farnazmdi:
Dr. Meyer, could you please see if this is correct? Nothing fails and I checked in the terminal that the functions in the class are working and nothing is breaking. If it is correct, JC and I can do this for the rest of the files.
Thank you.

### DIFF
--- a/lineage/CellVar.py
+++ b/lineage/CellVar.py
@@ -1,25 +1,17 @@
 """ This file contains the class for CellVar which holds the state and observation information in the hidden and observed trees respectively. """
 import numpy as np
-from typing import TypeVar, Generic
+from typing import TypeVar, Generic, Tuple
 from dataclasses import dataclass
 
-# temporary style guide:
-# Boolean functions are in camelCase.
-# Functions that return cells or lists of cells will be spaced with underscores.
-# Functions that are not to be used by a general user are prefixed with an underscore.
-# States of cells are 0-indexed and discrete (states start at 0 and are whole numbers).
-# Variables with the prefix num (i.e., num_states, num_cells) have an underscore following the 'num'.
-# Docstrings use """ and not '''.
-# Class names use camelCase.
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -15,3 +15,4 @@ jax==0.2.19
 jaxlib==0.1.70
 networkx==2.6.2
 statsmodels==0.12.2
+typing
```
aarmey, line 18: `typing` is built into Python. There is no module.

lineage/CellVar.py
```diff
@@ -150,14 +152,14 @@ def get_subtrees(node, lineage):
     return subtree, not_subtree
 
 
-def find_two_subtrees(cell, lineage):
+def find_two_subtrees(Cell: cell, lineage: list):
```
aarmey, line 155: I don't like having a type and variable of essentially the same name. Why not call the type `cellType`, and the argument `cell` like before?

lineage/CellVar.py
```diff
@@ -32,7 +34,7 @@ def __init__(self, parent, gen, **kwargs):
             self.right = kwargs.get("right", None)
             self.obs = kwargs.get("obs", [])
 
-    def divide(self, T):
+    def divide(self, T: np.ndarray):
```
aarmey, line 37:
```suggestion
    def divide(self, T: np.ndarray[float]):
```

aarmey (changes requested):
Close! A couple edits.

lineage/CellVar.py
Farnazmdi: It says `Type object is not subscriptable`..

aarmey: Ok, ignore that then.

Farnazmdi:
I am not authorized to merge. Could you please merge?

PR #851, opened 2021-10-07
Docstrings part 2

jclagarde:
My previous branch was too different from what was currently on master so I just created a new branch and a new PR. I was unsure if I correctly did the typing for the output of the `Analyze()` function and the `tHMMobj` parameter of the `Results()` function in Analyze.py however. Whenever you're free, could you take a look at those please?

### DIFF
--- a/lineage/Analyze.py
+++ b/lineage/Analyze.py
@@ -4,7 +4,8 @@ import numpy as np
 from concurrent.futures import ProcessPoolExecutor, Future, Executor
 from sklearn.metrics import balanced_accuracy_score
 from .tHMM import tHMM, fit_list
-from typing import Any, Tuple
+from typing import Any, Tuple, Union
+
 
 class DummyExecutor(Executor):
     def submit(self, fn, *args, **kwargs):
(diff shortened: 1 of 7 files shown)


Farnazmdi:
Please don't ask Dr. Meyer for a review yet. 
I am thinking we shouldn't need to import the typevar from a file to another if we want to use it. I am trying to find a way to make the type global so we don't need to import them. It will raise some errors because some functions are imported from file A to B, and some from B to A and it is a serious problem.

Farnazmdi:
yep! it should work.
Please ask Dr. Meyer for review/merge.
Thanks.

jclagarde:
Hi @aarmey, when you get the chance, please review the edits to the variable typing. If everything looks okay, please merge this branch. Thank you!

jclagarde:
(It's saturday night so feel free to respond to this at a time that is convenient for you)
In my recent branch, I edited the variable typing to pass the mypy tests. The only thing I wanted to ask if it was okay to do was in line 93 of Analyze.py, I had set the `results_dict` typing to be `results_dict: dict[Any, Any]` because it seems that mypy assumes all the values of dict to be of type `int`

jclagarde:
I just made one last small change to the typing for `results_dict` and it passes the mypy tests. Please let me know if there are any further changes I should make!

PR #853, opened 2021-10-14
docs added and updated

Farnazmdi:
This PR has two files in the Docs folder, `1.overview.rst` and `2.stateDistribution.rst` which I think they include enough information about getting started and using the package.
The problem is that mathematical expressions are not shown in mathematical expressions! Is is alright or we want to be sure that they show up in math? 
In a half-an-hour search, I couldn't find another way of showing math expression in .rst files. It seems it is a challenge for a lot of people...


Farnazmdi:
They do build now, A part of the overview looks like this:
image

PR #858, opened 2021-10-28
barcoding experiment

Farnazmdi:
Figure S16 (histogram) and figure S17 (Sankey plots)

I added a new property for cells (barcode) but since each lineage has a unique barcode, I didn't really need to use it, but I thought it wouldn't hurt to keep it as a cell's property.
I understand the histogram figures are not very elegant for going into a publication. If we are only showing them to the editors, they may suffice? If we are planning to insert them into the supp figures, please let me know.

I was not sure how to link a `go` figure to our `getSetup`. I created those figures in the terminal and saved them. I can manually upload them to the output folder if we want to have them in the repo. But the figure for both drugs is in the thmm channel in Slack.


aarmey (approved):
Great. I still need to think about how we'll respond to the reviewer comment, but we can do that when I review the reviewer responses.

Farnazmdi:
Thanks, could you please also merge this?

PR #864, opened 2021-11-12
fonts for letters and some other stuff

Farnazmdi:
Some small changes to the figure titles/legend. etc.
I had to build figure 4 because of the color of states 1 and 2 in the cartoon. I won't bore you with the details, but the order of colors when we plot the lineage tree needs to be re-organized only when we create figure 4 for consistency.


lineage/figures/figureCommon.py
```diff
@@ -361,7 +361,7 @@ def figureMaker(ax, x, paramEst, dictOut, paramTrues, xlabel="Number of Cells",
         ax[i].set_ylim(bottom=0, top=np.mean(tr) + 0.2)
         sns.regplot(x="x", y="tr", data=accuracy_df, ax=ax[i], lowess=True, marker='+', scatter_kws=scatter_kws_list[0])
         ax[i].set_ylabel(r"$||T-T_{est}||_{F}$")
-        ax[i].set_title(r"Error in estimating $T$")
+        ax[i].set_title(r"Error in estimating T")
```
aarmey, line 364: Look at capitalization here.

aarmey (changes requested):
Just saw one tiny edit. Did you mean to commit the figure?

Farnazmdi:
Yes, because of the color of states in the cartoon.

aarmey:
Merging!

PR #866, opened 2021-11-17
Poetry

aarmey:
This makes a few large structural changes to the repository:

- The environment is setup using poetry. This works exactly the same so long as you use the makefile commands.
- I removed the manuscript since the version is old.
- The figures to build is now automatically determined from the files that are available. Two figures files were renamed because they don't build.

@Farnazmdi please look at the built figures and check that nothing is broken. If they check out to you I'll merge.


Farnazmdi (approved):
Thank you!
The figures are building and they are looking good.

aarmey:
Great, thanks!

PR #877, opened 2021-12-24
Import the new Lineage Data

Farnazmdi:


Farnazmdi:
Right now, I am writing a test function for it, but probably won't work until I figure out the censored versus cell death part...

Farnazmdi:
Now that this is working, I guess need to write a new one! the data Dr. Dane sent is somehow different.

lineage/data/test.csv
```diff
@@ -0,0 +1,114 @@
+frame,trackId,lineageId,parentTrackId,Diameter_0
+0,1,2,0,10
```
aarmey, line 2: Why not test with the actual data?

aarmey (commented):
Excellent. One question here.

Note that this isn't MCF10A data. I believe this is tracking from the new AU565 experiments. The MCF10A data should be what Mark sent yesterday by email.

lineage/data/test.csv
Farnazmdi: At first, I did, but thought since I have to use a part of the code to put the cells from each lineage together, it might be better to make a synthetic one. If you don't agree, I will edit it and use the data itself.

Farnazmdi:
Oh.. Thanks for letting me know. I will work on importing the MCF10A data then.

lineage/data/test.csv
aarmey: I think it'd be better to import and test of the real data file. After that this should be ready to merge.

Farnazmdi: done.

aarmey (approved):
Excellent!

PR #878, opened 2021-12-31
MCF10A data import

Farnazmdi:


Farnazmdi:
EGF and HGF have some sort of error yet... I will have to investigate. Will do it tomorrow.

Farnazmdi:
@aarmey A problem in the data: apparently, one cell has 3 daughter cells in EGF treatment data... I am not sure what to do with this. Should I just ignore the third one?
image

aarmey:
Ignore the third one for now and then we can ask about this.

Farnazmdi:
There were some other data issues, I found another cell with three children in HGF treatment file, and another error was that one cell had 1 child.
I wrote some assertions and if-else statements.
One thing I don't know is the concentration of these drugs. Do we only have one concentration of each drug?

Farnazmdi:
At first I tried to use one function to import both cell lines, but they suffer from different data errors.. so I decided to separate them. I will work on a function that can import both cell lines later if it is alright.. I figured time is of the essence now!

aarmey:
> There were some other data issues, I found another cell with three children in HGF treatment file, and another error was that one cell had 1 child.

I'd suggest throwing out lineages with these problems right now, but print out a warning. We can ask about fixing these problems.

> I wrote some assertions and if-else statements.
> One thing I don't know is the concentration of these drugs. Do we only have one concentration of each drug?

These are growth factors. They were applied at only one concentration.

aarmey:
Completely agree with your choice to just keep these imports separate.

aarmey (approved):
Looks good.

PR #886, opened 2022-01-21
Order states

Farnazmdi:
This re-orders the states based on their gamma_shape value. I figured this parameter is the same for all conditions, so probably a consistent element to use for re-ordering.

Please let me know if you don't agree.


aarmey (approved):
Looks good. Merging.

Let's see if this leads to consistent results now.

PR #888, opened 2022-01-30
Run

Farnazmdi:
Sorting states proved not useful so I removed it. But I did see a repeating pattern. there is a cycle in 3 states and two other states converge to another state.
I had to update a couple of things, remade some figures and that's why there are many lines changed. The figures have been updated in dropbox.


aarmey:
Was sorting not helpful in that it worked but didn't help with interpretation, or it didn't work?

output/figure90.svg
```diff
@@ -0,0 +1,280 @@
+<?xml version="1.0" encoding="utf-8" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
+  "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
aarmey, line 3: Why is this committed? Isn't this going into one of the main figures?

output/figure91.svg
```diff
@@ -0,0 +1,1716 @@
+<?xml version='1.0' encoding='ASCII' standalone='yes'?>
+<svg xmlns:xlink="http://www.w3.org/1999/xlink" xmlns="http://www.w3.org/2000/svg" width="1144.679698pt" height="533.250619pt" viewBox="0 0 1144.679698 533.250619" version="1.1">
+ <metadata>
+  <rdf:RDF xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:cc="http://creativecommons.org/ns#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
```
aarmey, line 4: Why is this committed?

aarmey (changes requested):

output/figure90.svg
Farnazmdi: yes, I added it into figure8. If I run the model again the shape of it will probably be a little differentnt. If we are okay with not having the original figure here, I will remove it.

output/figure91.svg
Farnazmdi: I will delete this one.

Farnazmdi:
The sorting function was based on the shape parameter that is shared among all conditions, but at every run, the maximum shape was not corresponding to the same state and it was changing. I could not figure out another good criterion to sort states based on it, and I did see a pattern repeating, so I removed it.

aarmey:
Ok, we can merge this without it, but could you move some of the code (in a separate PR) to a function that takes in a permutation (e.g., [1, 0, 3, 2]) and reorganizes the tHMM model. Basically this is the function you have, but without the rule on how to rearrange them. Some of this is also done by `permute_states()` in `Analyze()`, and so having this function would (1) remove some duplication, and (2) allow us to easily add a rule for permuting later.

Farnazmdi:
Absolutely. Thank you.

PR #889, opened 2022-01-31
edited permute states function

Farnazmdi:
I had not noticed that we have the permute_states function.
So I just made the edits you mentioned in the other PR. This function takes in the tHMMobject and a mapping array to rearrange the states.


aarmey (commented):
Aren't the removed lines still necessary to align the benchmark fitting? As in they can be moved out of the function but should still be used in `Results()`? I know that we no longer have to worry about the state ordering for the state assignment accuracy, but I think we still need them aligned to quantify the accuracy of the emissions distributions...

Farnazmdi:
Dne.

aarmey (approved):

PR #906, opened 2022-04-26
reviewer comments

Farnazmdi:
1. figure6.py updated, baseline classifier instead of kmeans.
In the next comment I wrote the mathematical expression I used for the threshold for the classifier.
2. changed "Random Index accuracy" to "Adjusted Rand Index" in all figures (I will ask JC to change this in the figures that we won't re-make).


Farnazmdi:
image
image

lineage/states/stateCommon.py
```diff
@@ -18,7 +18,7 @@
 def nLL_sep(x, uncens_obs, uncens_gammas, cens_obs, cens_gammas):
     a, scale = x
     uncens = jnp.dot(uncens_gammas, gamma.logpdf(uncens_obs, a=a, scale=scale))
-    cens = jnp.dot(cens_gammas, gammaincc(a, cens_obs / scale))
+    cens = jnp.dot(cens_gammas, jnp.log(gammaincc(a, cens_obs / scale)))
```
aarmey, line 21: Are you absolutely sure you need this change? If not, remove it.

Farnazmdi: The upper one is in the log form, and the lower one was not. And we are summing them together. I think we should keep it this way.

aarmey: This is not an acceptable justification for changing the equation. You need to look back at the paper Shakthi found for this and confirm the log is needed. Alternatively, if you see it definitely improves the fitting results, I might be convinced it's correct based on that alone.

aarmey: Either way I'd suggest removing this from the current PR, because the rest of the changes generally look good.

Farnazmdi: removed.

lineage/figures/common.py
```diff
@@ -328,15 +328,10 @@ def commonAnalyze(list_of_populations, num_states, xtype="length", **kwargs):
     list_of_fpi = kwargs.get("list_of_fpi", [None] * len(list_of_populations))
     list_of_fT = kwargs.get("list_of_fT", [None] * len(list_of_populations))
     list_of_fE = kwargs.get("list_of_fE", [None] * len(list_of_populations))
-    if num_states == 2:
-        predicted_num_states = kwargs.get("predicted_num_states", 2)
-    elif num_states == 3:
```
aarmey, line 333: Was this here for some reason before? I can see getting rid of this code, but I can't understand why it was written this way in the first place...

lineage/figures/figure6.py
```diff
@@ -91,9 +96,20 @@ def figureMaker5(ax, distribution_df, wasser_df):
...
     ax[i].legend()
+
+def classification_threshold(shape, scale1, scale2):
+    """ Given the parameters of the gamma distribution, it provides an analytical threshold for classification.
+    This function is specific to this figure, as the shape parameter is shared and only the scale varies.
+    """
+    if scale1 == scale2:
+        return shape * scale1
+    else:
+        numer = shape * np.log(scale2/scale1)
```
aarmey, line 113: What is your reasoning behind this threshold?

My thinking was that we might even be able to get rid of using the data altogether, and just draw the accuracy line from an analytical expression. Maybe that's too ambitious...

aarmey: Oh! I see above that you're finding the point at which they are equal probability. Clever. I'm fine with that.

aarmey (approved):
One question remaining here, but this looks good.

lineage/figures/common.py
Farnazmdi: `predicted_num_states` has not been referenced anywhere else, so I think there was no point having it. I am not sure why this was written like this.

lineage/figures/figure6.py
Farnazmdi: when scale1 == scale2, the denominator becomes zero, and the two distributions are exactly the same, so I assumed the mean of the distribution would be the threshold which I think should result in 50% accuracy.
If you have any other suggestions I can edit this.

I think I can do that by working with the areas under the pdfs. I will work on it.

aarmey: But the scale isn't the same—the shape is the same.

aarmey: Oh I see. That's your fallback. That's fine.

PR #910, opened 2022-04-26
transition matrix not shared

Farnazmdi:
This is the PR with changes for un-sharing transition matrix.


Farnazmdi:
@aarmey I believe this PR is ready for merging. I have found reasonable similarities between this run and previous results and will talk about them on Wednesday. I didn't build any new figures in the output folder, just updated those that were already there.

lineage/BaumWelch.py
```diff
@@ -88,9 +88,9 @@ def do_M_step(tHMMobj, MSD: list, betas: list, gammas: list):
         assert tHMMobj[0].fT is None
         T = do_M_T_step(tHMMobj, MSD, betas, gammas)
 
-        # all the objects in the population have the same T
-        for t in tHMMobj:
-            t.estimate.T = T
+        # each tHMMobj has its own T
+        for jj, t in enumerate(tHMMobj):
+            t.estimate.T = T[jj]
```
aarmey, line 93: This changes the transition matrix setup for both drugs. I thought we're trying to find support for sharing the transition matrix across the drug concentrations, as it is currently done?

Farnazmdi: oh... I thought we do want to settle on having T separate for each condition.
Either way, we can use the figures I made for the response to reviewers. I will revert this code change.
So technically we don't want these changes... I will only keep the changes you made to fix that error.

Farnazmdi: Should be done. The changes I kept are necessary (some are from simplifying `Results()` that I had not noticed need change in the other PR)
Should be ready to merge..

lineage/figures/common.py
```diff
@@ -612,8 +612,8 @@ def plot_all(ax, num_states, tHMMobj_list, Dname, cons, concsValues):
     print(Dname, "\n the \u03C0: ", tHMMobj_list[0].estimate.pi, "\n the transition matrix: ", tHMMobj_list[0].estimate.T)
     for idx, tHMMobj in enumerate(tHMMobj_list):  # for each concentration data
         for i in range(num_states):
-            lpt_avg[idx, i, 0] = tHMMobj.estimate.E[i].params[2] * tHMMobj.estimate.E[i].params[3]  # G1
-            lpt_avg[idx, i, 1] = tHMMobj.estimate.E[i].params[4] * tHMMobj.estimate.E[i].params[5]  # G2
+            lpt_avg[idx, i, 0] = np.log10(tHMMobj.estimate.E[i].params[2] * tHMMobj.estimate.E[i].params[3])  # G1
```
aarmey, line 615: Does this affect the figures?

aarmey (approved):
One more question. Otherwise, this looks good.

lineage/figures/common.py
Farnazmdi: no, The figures are at the scale of log10, but I am not sure why this had been removed.

PR #919, opened 2022-05-10
mix of figure edits and cv

Farnazmdi:


Farnazmdi:
Note:
The "atonce fitting" (`trust-constr`) fails sometimes and that's when the estimated parameters are very small and all equal like this [0.11510162 0.1151016  0.11510161 0.11510161 0.11510162]
I need to write an `if` so in these cases change the x to some good initial point and run minimization again.

Farnazmdi:
Note2: I made `keep_feasible=False` since the estimator was throwing errors occasionally when the estimated parameter was outside the bound and it was failing often.

PR #931, opened 2022-05-31
Fix censored estimator

aarmey:
You were right—the distribution estimator was wrong. The survival function, which should be used for censored observations, correctly should be `gammaincc`. However, our optimization is applied with log-likelihoods, and so it should have been `log(gammaincc)`. I changed this to use scipy's distribution to avoid misunderstandings in the future.


aarmey:
Nvm. I'm doing it.

PR #940, opened 2022-06-03
cross validation additions

Farnazmdi:
This branch is purely edits related to cross validation.


lineage/states/StateDistributionGamma.py
```diff
@@ -68,7 +68,7 @@ def pdf(self, x: np.ndarray):
         # Update for observed Bernoulli
         ll[np.isfinite(x[:, 0])] += sp.bernoulli.logpmf(x[np.isfinite(x[:, 0]), 0], self.params[0])
 
-        # Make sure censored values are censored
+        # likelihood of negative values should be zero
```
aarmey, line 71: This is a log-likelihood and should be labelled as such.

lineage/states/StateDistributionGamma.py
```diff
@@ -68,7 +68,7 @@ def pdf(self, x: np.ndarray):
...
         ll[x[:, 1] < 0] = 0.0
 
         return np.exp(ll)
```
aarmey, line 74: It might be useful to make a logpdf function in this class, since you probably want that most of the time. The pdf method can just call logpdf then take the exp.

lineage/crossval.py
```diff
@@ -1,15 +1,64 @@
...
+                for cell_indx, cell in enumerate(lin.output_lineage):
+                    cell.state = states_list[k][idx][lin_indx][cell_indx]
+
+        # calculate the likelihood of hidden observations to their predicted state
+        for idx, tHMMobj in enumerate(tHMMobj_list):
+            for lin_indx, lin in enumerate(tHMMobj.X):
+                for cell_indx, cell in enumerate(lin.output_lineage):
+                    if cell.obs[2] < 0:
+                        positive_obs = [-1 * o for o in cell.obs]
+                        Ls += tHMMobj.estimate.E[cell.state].pdf(np.array(positive_obs)[np.newaxis, :])
```
aarmey, line 61: You do not want to add these if they're likelihoods. Only want to do that if they're log-likelihoods (they're not).

lineage/crossval.py
```diff
@@ -1,15 +1,64 @@
...
+    tHMMobj_list_states = []
+    for out in output:
+        tHMMobj_list_states.append(out[0])
+
+    # predict states of hidden cells
+    # states_list: len(states_list) = num_states. states_list[0] for the 1-state model, etc.
+    states_list = [[tHMMobj.predict() for tHMMobj in tHMMobj_list] for tHMMobj_list in tHMMobj_list_states]
+
+    # calculate the likelihood of observations of masked cells to their assigned state
+    LLs = []
```
aarmey, line 44: You need to be very careful about naming for log-likelihoods vs. likelihoods.

lineage/crossval.py
```diff
@@ -1,15 +1,64 @@
...
 
 def hide_observation(lineages: list, percentage: float) -> list:
     """ Taking a list of lineages and the percentage of cells want to be masked, it marks those x% negative."""
     new_lineages = deepcopy(lineages)
     for new_lineage in new_lineages:
-        for cell in new_lineage.output_lineage:
-            if bernoulli.rvs(p=percentage, size=1):
-                # negate the cell observations to mask them
-                cell.obs = [-1 * o for o in cell.obs]
+        if len(new_lineage.output_lineage) < 2:
```
aarmey, line 12: Why can't you censor a lineage with one cell?

lineage/tests/test_CrossVal.py
```diff
@@ -18,3 +21,16 @@ def test_hide_obs():
                 negatives += 1
 
     assert 0.2 <= negatives / total <= 0.3
+
+@pytest.mark.parametrize("cen", [0, 3])
+def test_cv(cen):
+    complete_lineages = [[LineageTree.init_from_parameters(pi, T, E2, 7, censored_condition=cen, desired_experiment_time=100) for _ in range(50)] for _ in range(4)]
```
aarmey, line 27: Docstring.

lineage/crossval.py
```diff
@@ -1,15 +1,64 @@
...
+            continue
+        else:
+            for cell in new_lineage.output_lineage:
+                if bernoulli.rvs(p=percentage, size=1):
+                    # negate the cell observations to mask them
+                    cell.obs = np.array([-1 * o for o in cell.obs])
```
aarmey, line 18: I'm not sure you want to change the type of `cell.obs` from a list to an ndarray only when masked. However, I do agree it should be an ndarray. Perhaps you can make this `cell.obs *= -1` and then find where it's being defined as a list and convert those spots to an ndarray?

lineage/crossval.py
aarmey: Here is exactly where you want to loop over states, too, so just make the change here.

aarmey (changes requested):
Some things need to be cleared up here.

lineage/crossval.py
Farnazmdi: my bad, it should and it is fixed.

lineage/crossval.py
Farnazmdi: I tried to do that, but it messed with different things. If it is okay, I will keep the cell.obs as a list everywhere.

lineage/crossval.py
```diff
@@ -13,3 +15,46 @@ def hide_observation(lineages: list, percentage: float) -> list:
...
+                    cell.state = states_list[k][idx][lin_indx][cell_indx]
+
+        Logls = 0
+        # calculate the log likelihood of hidden observations
+        for idx, tHMMobj in enumerate(tHMMobj_list):
+            for lin_indx, lin in enumerate(tHMMobj.X):
+                for cell_indx, cell in enumerate(lin.output_lineage):
+                    if cell.obs[2] < 0:
+                        positive_obs = [-1 * o for o in cell.obs]
+                        Logls += tHMMobj.estimate.E[cell.state].logpdf(np.array(positive_obs)[np.newaxis, :])
```
aarmey, line 57: This is wrong. The term you should be adding to the likelihood should come from making an array of `gamma[i] + logpdf(emission)` where i is each state. You do not need `states_list`, you need the soft assignments as given by `gamma`. These likelihoods for one cell across states should then be added together using `logsumexp`, then can be added to the overall log-likelihood.

aarmey (changes requested):
This is very close, but should be correct before we merge.

lineage/crossval.py
aarmey: Yep, fine to keep it this way.

lineage/crossval.py
Farnazmdi: why should we add the likelihoods for one cell across states together? Shouldn't we keep the `max` of the values we get from  `gamma[i] + logpdf(emission)`?

aarmey: We have talked about this several times before. The model does not provide hard assignment of cells to one state in particular. You have to consider the relative amounts of assignment to each state.

Farnazmdi: apologies, it *should* be correct now.

Farnazmdi: image
sounds promising.

aarmey (approved):
Looks good!

aarmey:
@Farnazmdi let me know how this works with the experimental data.

PR #944, opened 2022-06-08
leave out part of cross validation

Farnazmdi:
1. The estimators are updated such that they ignore negative observations.
2. During the calculation of pdf of observations, those cells with negative observation will get an equal probability to each state, this way they do not contribute to fitting.

### DIFF
--- a/lineage/states/StateDistributionGamma.py
+++ b/lineage/states/StateDistributionGamma.py
@@ -68,7 +68,7 @@ class StateDistribution:
         # Update for observed Bernoulli
         ll[np.isfinite(x[:, 0])] += sp.bernoulli.logpmf(x[np.isfinite(x[:, 0]), 0], self.params[0])
 
-        ll[x[:, 0] < 0] = 0.0
+        ll[x[:, 1] < 0] = 0.0
 
         return np.exp(ll)
 
--- a/lineage/states/StateDistributionGamma.py
+++ b/lineage/states/StateDistributionGamma.py
@@ -45,7 +45,7 @@ class StateDistribution:
         """
         return 3
 
-    def pdf(self, x: np.ndarray):
+    def pdf(self, x: np.ndarray, num_states=2):
         """ User-defined way of calculating the likelihood of the observation stored in a cell.
         In the case of a univariate observation, the user still has to define how the likelihood is calculated,
         but has the ability to just return the output of a known scipy.stats.<distribution>.<{pdf,pmf}> function.
@@ -67,6 +67,8 @@ class StateDistribution:
 
         # Update for observed Bernoulli
         ll[np.isfinite(x[:, 0])] += sp.bernoulli.logpmf(x[np.isfinite(x[:, 0]), 0], self.params[0])
+
+        ll[x[:, 0] == -1] = np.log(1/num_states)
 
         return np.exp(ll)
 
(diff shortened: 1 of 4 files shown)


lineage/states/StateDistributionGamma.py
```diff
@@ -68,6 +68,8 @@ def pdf(self, x: np.ndarray):
         # Update for observed Bernoulli
         ll[np.isfinite(x[:, 0])] += sp.bernoulli.logpmf(x[np.isfinite(x[:, 0]), 0], self.params[0])
 
+        ll[x[:, 0] == -1] = np.log(1/num_states)
```
aarmey, line 71: Why do you need this? It seems like you should be able to just set the log-likelihood to 0.

lineage/states/stateCommon.py
```diff
@@ -79,20 +79,25 @@ def gamma_estimator(gamma_obs: list[np.ndarray], time_cen: list[np.ndarray], gam
...
 
-    if len(gamma_obs) == 4:  # for constrained optimization
+    if len(gamma_obs_) == 4:  # for constrained optimization
         A = np.zeros((3, 5))  # is a matrix that contains the constraints. the number of rows shows the number of linear constraints.
         np.fill_diagonal(A[:, 1:], -1.0)
         np.fill_diagonal(A[:, 2:], 1.0)
         linc = [LinearConstraint(A, lb=np.zeros(3), ub=np.full(3, np.inf))]
         if np.allclose(np.dot(A, x0), 0.0):
-            x0 = np.array([200.0, 0.2, 0.4, 0.6, 0.8])
+            x0 = np.array([200.0, 0.5, 0.2, 0.4, 0.6, 0.8])
```
aarmey, line 100: Why is there a new parameter here? Or has it been wrong?

aarmey (changes requested):
Generally good. Just a couple questions.

lineage/states/StateDistributionGamma.py
Farnazmdi: It was my first intuition, but putting the likelihood of those cells to zero will create some problems in the later steps that something will be divided by this value and threw an error.

lineage/states/stateCommon.py
Farnazmdi: oops, my bad. it was right, I will fix it now.

lineage/states/StateDistributionGamma.py
Farnazmdi: I have not tried setting it to a small number like epsilon though. I will try that and see if it still errors.

aarmey: This is more reason to set it to 0! Under no circumstances should the code be dividing by the log-likelihood. If a 0 causes an error here, that indicates a problem that needs to be fixed elsewhere.

Farnazmdi: I think your recent changes to the estimators fixed that. I don't get that error anymore when I set the log-likelihood to zero, but get more cost function = inf.
I am trying to find a way to make the `minimize` run again if the cost function becomes inf.

aarmey: Hmm... where does inf first arise? Is it in the estimator? It's likely a quick fix if we can track it down.

Farnazmdi: It is inside the `gamma_LL()` line 51 stateCommon.py (where you have `if np.isinf(outt): print(x)`). The assersion about `outt` being finite is violated.

lineage/states/stateCommon.py
```diff
@@ -79,14 +79,19 @@ def gamma_estimator(gamma_obs: list[np.ndarray], time_cen: list[np.ndarray], gam
     if np.sum([np.sum(g) for g in gammas]) < 0.1:
         gammas = [np.ones(g.size) for g in gammas]
 
+    # make sure all negative observations are removed
+    gamma_obs_ = [gm[gamma_obs[i] >= 0] for i, gm in enumerate(gamma_obs)]
```
aarmey, line 83: Does the estimator need to do this step, if it's already done in the state distribution class? It seems like that can just pass the filtered variables...

aarmey (changes requested):

lineage/states/stateCommon.py
Farnazmdi: done.

aarmey (approved):
Great. Merging!

PR #945, opened 2022-06-08
hide observations in a list of lineages

Farnazmdi:
1. Given a list of lineages and a percentage, masks that percentage of all cells within that list of lineages, and return the new lineages where the observations of masked cells are marked as -1. Also returns the indexes of hidden states and those hidden observations for later use. 
2. a unit test that makes sure the cells' observations are actually marked as -1.

### DIFF
--- a/lineage/crossval.py
+++ b/lineage/crossval.py
@@ -34,16 +34,13 @@ def hide_observation(lineages: list, percentage: float) -> Tuple[list, list, lis
         new_hide_index.append(hide_index[prev:prev+i])
         prev += i
 
-    obss = [] # save those observations that will be masked
     for i, new_lineage in enumerate(new_lineages):
         tmp1 = []
         for ix, cell in enumerate(new_lineage.output_lineage):
             if new_hide_index[i][ix] == 1: # means we hide the cell lifetime
                 tmp1.append(cell.obs)
-                cell.obs = -1 * np.ones(len(cell.obs))
-        obss.append(tmp1)
+                # negate the cell observations at those specific indexes
+                cell.obs = [-1 * o for o in cell.obs]
 
-    for i, ob in enumerate(obss):
-        assert np.sum(new_hide_index[i]) == len(ob)
 
-    return new_lineages, new_hide_index, obss
+    return new_lineages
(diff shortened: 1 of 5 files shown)


lineage/crossval.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # to partition the hide_index (which is an array that has all cells in all lineages together) 
+    # each array for each lineage
+    prev = 0
+    new_hide_index = []
+    for i in len_lineage:
+        new_hide_index.append(hide_index[prev:prev+i])
+        prev += i
+
+    obss = [] # save those observations that will be masked
```
aarmey, line 37: Would it be cleaner to not duplicate the observations? They are still stored within the lineage, since they're present but just negative.

lineage/crossval.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # create the indexes for hidden observations
+    hide_index = np.zeros(num_cells)
+    hide_index[:int(num_cells*percentage)] = 1
+    hide_index = shuffle(hide_index)
+
+    # to partition the hide_index (which is an array that has all cells in all lineages together) 
+    # each array for each lineage
+    prev = 0
+    new_hide_index = []
```
aarmey, line 32: This is also possible to find by inspecting the held-out tree, too, right?

aarmey (changes requested):

lineage/crossval.py
Farnazmdi: sure, it makes more sense. will do.

lineage/crossval.py
Farnazmdi: yes, will do.

lineage/crossval.py
```diff
@@ -0,0 +1,45 @@
...
+    num_cells = 0
+    len_lineage = [] # remember the length of each lineage
+    for lin in lineages:
+        num_cells += len(lin.output_lineage)
+        len_lineage.append(len(lin.output_lineage))
+
+    # create the indexes for hidden observations
+    hide_index = np.zeros(num_cells)
+    hide_index[:int(num_cells*percentage)] = 1
+    hide_index = shuffle(hide_index)
```
aarmey, line 26: One more thought—this doesn't need to be an exact fraction. Why not just walk the cell tree, and then pull a Bernoulli random number for each cell and flip it if you get a 1? That further massively simplifies all of this.

Farnazmdi: That certainly made it a lot smpler! Thank you!

PR #949, opened 2022-06-11
cross validation on experimental data

Farnazmdi:
two runs for experimental data:
image
image
Maybe lapatinib 4 and gemcitabine 5? I had hoped it would plateau in smaller states..


aarmey (approved):
Looks great! The result is what it is, but this is a much stronger basis on which to claim we find multiple states (and the model is a good representation of the data).

PR #952, opened 2022-06-21
all figures updated

Farnazmdi:
This branch just updated all figures according to the new fitting. The only thing that I "changed" was that in the BIC figure, we were fitting and calculating for 1, 2, .., 7 states and pick the best number then run the model again with those number of states. What I did in this branch is that during the BIC fitting, I save the tHMMobjects and recall the one that we pick based on the BIC. This way the BIC value matches the tHMMobj we pick.


aarmey:
@Farnazmdi this is failing because `cell.state` is None. Did you possibly pickle the tHMM objects before they were fit?

Farnazmdi:
Sorry about this. I will fix it tonight. 
I saved all models created from BIC analysis (1, 2, 3, .. state tHMM objects) so we don't have to fit the model again with the optimal number of states. I have not assigned states to those tHMMobjs created from BIC fitting and pickled. 
This is failing probably in one or two figures that I forgot to add the state assignment code.

aarmey:
That's reasonable. Just trying to help with where the problem lies.

I also think it's better to not commit these figure files. The pickle files now keep the results reproducible.

Farnazmdi:
I removed all svg files that could be recreated. Had to make some changes to the functions to automatically generate those figures in the `cartoon` folder. 
This should be ready now. I will merge it if it is alright? The changes are mostly just to the figures, with a minor change to the BW test and the addition of a function to plotTree to create those samples of lineage trees we use in lapatinib and gemcitabine figures.

aarmey:
Let me go through it carefully first. Will do so in a bit.

aarmey (approved):
Ok, all looks good!

Farnazmdi:
Thank you, merging..

PR #964, opened 2022-08-17
Generations and parents by index array

aarmey:


aarmey:
This one's quite a bit of work to follow. It removes many of the `list.index()` operations for speed. Potentially some of these functions could now be wrapped with numba to speed them up later, too.

Farnazmdi (approved):
Thank you so much, it is much more readable. Do you mind if I add some docstring to remember?

PR #978, opened 2023-02-07
removed pickle files

Farnazmdi:
This PR removes all pickle files and codes associated with them.


aarmey (approved):
Excellent, thanks! I'll try and open a PR with some of my initial changes.

PR #981, opened 2023-02-09
set up random seed for reproducible results

Farnazmdi:
In this PR:
- Added random init seed for state distributions (E) and when we initialize the model parameters (T,pi).
The tests still occasionally fails..

### DIFF
--- a/lineage/Analyze.py
+++ b/lineage/Analyze.py
@@ -103,14 +103,14 @@ def Analyze_list(
     tHMMobj_list = [
         tHMM(X, num_states=num_states, **kwargs) for X in pop_list
     ]  # build the tHMM class with X
-    rng = kwargs.get('rng', None)
+    rng = kwargs.get('rng', 123)
     _, _, _, gammas, LL = fit_list(tHMMobj_list, rng)
 
     for _ in range(5):
         tHMMobj_list2 = [
             tHMM(X, num_states=num_states, **kwargs) for X in pop_list
         ]  # build the tHMM class with X
-        _, _, _, gammas2, LL2 = fit_list(tHMMobj_list2)
+        _, _, _, gammas2, LL2 = fit_list(tHMMobj_list2, rng)
 
         if LL2 > LL:
             tHMMobj_list = tHMMobj_list2
--- a/lineage/Analyze.py
+++ b/lineage/Analyze.py
@@ -57,7 +57,7 @@ def fit_list(
     # when there are no fixed emissions, we need to randomize the start
     init_gam = [
         [
-            sp.dirichlet.rvs(np.ones(tO.num_states), size=len(lin), random_state=rng)
+            rng.dirichlet(np.ones(tO.num_states), size=len(lin))
             for lin in tO.X
         ]
         for tO in tHMMobj_list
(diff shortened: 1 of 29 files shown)


aarmey:
The test needs to initialize and then pass in the rng to ensure it is set to the same state. I am not seeing anything set within the tests that are failing...

lineage/Analyze.py
```diff
@@ -103,13 +103,14 @@ def Analyze_list(
...
 
     for _ in range(5):
         tHMMobj_list2 = [
             tHMM(X, num_states=num_states, **kwargs) for X in pop_list
         ]  # build the tHMM class with X
-        _, _, _, gammas2, LL2 = fit_list(tHMMobj_list2)
+        _, _, _, gammas2, LL2 = fit_list(tHMMobj_list2, rng)
 
         if LL2 > LL:
             tHMMobj_list = tHMMobj_list2
```
Farnazmdi, line 116: @aarmey In this function, we run the fitting multiple times and choose the best fit. If we have the random seed here, this multiple running would be meaningless, and also, if we don't set a random seed and run this fitting multiple times, then there is no guarantee it finds the same answers.. 
Should we have only 1-time fitting here?

aarmey: If you are passing the `rng` around, the seed will be different each fit, but the collection of them will be reproducible. e.g., the first time it'll be 1, second time 6, third time 23, etc.

lineage/tHMM.py
```diff
@@ -19,19 +20,20 @@ def __init__(self, X: list[LineageTree], nState: int, fpi=None, fT=None, fE=None
...
         """
+
         self.fpi = fpi
         self.fT = fT
         self.fE = fE
         self.num_states = nState
 
         if self.fpi is None or self.fpi is True:
-            self.pi = np.random.rand(nState)
+            self.pi = sp.uniform.rvs(0.0, 1.0, nState)
```
aarmey, line 30: This isn't using the random number generator. You can use [`rng.uniform()`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.uniform.html) here.

lineage/tHMM.py
```diff
@@ -19,19 +20,20 @@ def __init__(self, X: list[LineageTree], nState: int, fpi=None, fT=None, fE=None
...
             self.pi /= np.sum(self.pi)
         else:
             self.pi = self.fpi
 
         if self.fT is None:
-            self.T = np.random.dirichlet(np.random.rand(nState), nState)
+            self.T = sp.dirichlet.rvs(np.ones(nState), size=nState, random_state=rng)
```
aarmey, line 36: [`rng.dirichlet()`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.dirichlet.html) doesn't need scipy.

aarmey:
@Farnazmdi this should now be working. Please look this over then I can merge.

Farnazmdi:
Thank you so much..
Just one question. Does it matter that we use the same random seed number in all files? Or it just matters that those numbers, in any function, stay the same when we run the functions at different times?

aarmey:
I would suggest you read through the numpy documentation, and think about how this works within the context of the analysis. This is something you should be able to answer at this point.

